PR #113, opened 2026-05-12
feat(mail): add draft composition command

rianjs:
## Summary

Adds `gro mail draft` — composes a Gmail draft via `users.drafts.create`. Drafts land in the user's Drafts folder for human review; the CLI never calls `drafts.send`, preserving gro's non-destructive philosophy.

Body input is **markdown by default**, rendered to HTML via goldmark (table, strikethrough, task-list extensions). `--plain` and `--html` opt out of rendering. Body comes from `--body`, `--stdin`, or `--file`. Recipients are comma-separated address lists.

## Example usage

```bash
# Markdown rendered to HTML (default)
gro mail draft --to alice@example.com --subject "Hi" --body "**hello**"

# Body from file or stdin
gro mail draft --to a@x.com --subject "Report" --file report.md
echo "# Status" | gro mail draft --to a@x.com --subject "Update" --stdin

# Plain text or raw HTML
gro mail draft --to a@x.com --subject "Plain" --body "no md" --plain
gro mail draft --to a@x.com --subject "Raw" --body "<h1>Hi</h1>" --html

# Attachments + send-as alias
gro mail draft --from work@me.com --to a@x.com --subject "Hi" --body "see attached" --attach report.pdf
```

## MIME assembly

The gmail package owns MIME construction and acts as the security boundary:

- RFC 5322/2045 compliant: `MIME-Version: 1.0`, CRLF line endings, `Content-Transfer-Encoding: base64` with 76-column wrapping.
- Single-part for plain bodies, `multipart/mixed` for attachments.
- Subject is RFC 2047 encoded for non-ASCII; attachment filenames RFC 2231 encoded.
- **CR/LF header-injection guard** independent of command-layer validation.
- Attachment filenames are `filepath.Base`-stripped — local paths never leak.

## Coverage

- 21 MIME-builder + httptest API-wiring tests in `internal/gmail/drafts_test.go`
- 21 command-level tests in `internal/cmd/mail/draft_test.go` (flag presence, validation negatives including CR/LF injection, all success paths)
- 18 integration tests run against the live Gmail API — all green. See `integration-tests.md`.

## Scope discipline

Out of scope (intentionally): no `drafts.send`, `drafts.update`, `drafts.list`, `drafts.delete`; no reply threading; no `$EDITOR` fallback; no `--dry-run` (creating a draft is itself the safe preview).

Closes #112

## Test plan

- [x] All unit tests pass (`make test` — 1291 tests across 28 packages)
[...]

PR #115, opened 2026-05-13
feat(mail): reply-to-thread drafts

### LINKED ISSUE #114: feat(mail): reply-to-thread drafts
## Summary

Extend the \`gro mail draft\` command (added in #112 / #113) so it can compose a reply that Gmail (and any standards-compliant MUA) renders as a real reply: grouped under the original thread, with \`In-Reply-To\` and \`References\` headers pointing at the parent message.

The non-destructive philosophy carries over verbatim — the CLI still never calls \`drafts.send\`. The draft sits in the user's Drafts folder for explicit human review.

## Reuse, don't rebuild

This is **strictly additive** on top of the #112 / #113 work. Specifically reuse:

- \`internal/gmail/drafts.go\` → \`DraftMessage\`, \`DraftAttachment\`, \`DraftResult\`, \`buildMIME\`, \`CreateDraft\`. Add fields to \`DraftMessage\`; do not introduce a parallel \`ReplyMessage\` type.
- The full MIME assembly pipeline (\`MIME-Version: 1.0\`, CRLF, base64 wrap, RFC 2047 subject encode, RFC 2231 filenames, multipart/mixed for attachments). Reply mode adds **two more headers** to the existing builder, nothing else.
- The header-injection guard in \`guardHeaderInjection\` — extend it to cover the new \`InReplyTo\` and \`References\` fields rather than writing a new guard.
- \`internal/cmd/mail/draft.go\` → \`newDraftCommand\`. Add two new flags to the existing command; **do not** create a separate \`gro mail reply\` command.
- The \`MailClient\` interface and \`MockGmailClient\` in \`internal/cmd/mail/output.go\` / \`mock_test.go\` — \`CreateDraft\` already handles arbitrary \`DraftMessage\` shapes, so no interface change is needed for the API call itself. Only add a new mock entry if we need it for the source-message lookup (which already exists as \`GetMessage\`, also already on the interface).
- The markdown rendering, plain/HTML mode flags, attachment handling, comma-separated address parsing, \`--from\` normalisation, and \`cobra.NoArgs\` — none of those need to change.
- The integration-tests.md draft section — append T21+ scenarios alongside the existing T1-T20.

If you find yourself duplicating any of these, stop and refactor to share instead.

## Background

Today \`gro mail draft\` always creates a new standalone thread because:
- \`DraftMessage\` has no \`ThreadID\`, \`InReplyTo\`, or \`References\` fields.
- The Gmail API call doesn't set \`Draft.Message.ThreadId\`, so the server assigns a fresh thread.
- No \`In-Reply-To\`/\`References\` MIME headers are emitted, so even when forced into the right thread, downstream MUAs may not show the reply indicator.

## Proposed usage

```bash
# Simplest form — let gro look up the source message
gro mail draft --reply-to <message-id> --body \"thanks, will review\"

# Overrides still work — recipients, subject, body source, attachments, --from
gro mail draft --reply-to <message-id> --cc colleague@example.com --body \"...\"
gro mail draft --reply-to <message-id> --subject \"Re: customised\" --body \"...\"

# Reply-all behaviour
gro mail draft --reply-to <message-id> --reply-all --body \"...\"
```

Without \`--reply-to\`, behaviour is unchanged from #113. With \`--reply-to\` and no overrides:
- To = original \`From\`
- Cc = (empty)
- Subject = \`Re: <orig subject>\` if not already prefixed
- \`In-Reply-To\` = original \`Message-Id\`
- \`References\` = original \`References\` (if any) + original \`Message-Id\`
- \`Draft.Message.ThreadId\` = original thread ID

With \`--reply-all\`:
- To = original \`From\`
- Cc = original \`To\` + original \`Cc\`, minus the authenticated user

Explicit \`--to\`/\`--cc\`/\`--bcc\`/\`--subject\` override the derived values (use \`cmd.Flags().Changed()\`, matching the existing pattern in #113).

## API surface (no scope change)

- \`gmail.GmailModifyScope\` already covers reading message headers and creating drafts. No \`auth.AllScopes\` change.
- Reuses existing \`(*gmail.Client).GetMessage(ctx, id, includeBody=false)\` to fetch headers — already on the \`MailClient\` interface.

### \`internal/gmail/drafts.go\`
Add three fields to the existing \`DraftMessage\`:
\`\`\`go
type DraftMessage struct {
    // ... existing fields from #113 ...
    ThreadID   string   // empty = new thread; set = reply
    InReplyTo  string   // original Message-Id (with angle brackets)
    References []string // accumulated chain
}
\`\`\`

Extend the existing \`buildMIME\` to emit \`In-Reply-To\` and \`References\` headers when set. Extend \`guardHeaderInjection\` to cover these two fields. Extend the existing \`CreateDraft\` body to set \`Draft.Message.ThreadId = msg.ThreadID\` before the API call.

### \`internal/cmd/mail/draft.go\`
Add two flags to the existing \`newDraftCommand\`:
- \`--reply-to <message-id>\` (string)
- \`--reply-all\` (bool, only meaningful with \`--reply-to\`)

\`RunE\` gains a resolution step before existing validation:
1. If \`--reply-to\` is set, call \`client.GetMessage(ctx, replyToID, false)\` to read headers.
2. Derive defaults from the source message.
3. Explicit flags override derived values (cobra's \`Changed()\` — same pattern used for \`--subject\` and \`--file\` today).
4. Subject auto-prefix with \`Re: \` if not already present (case-insensitive check).

### Tests
- gmail layer: extend the existing tables in \`drafts_test.go\` — \`In-Reply-To\`/\`References\` emitted in MIME, parsed back via \`net/mail\`; injection guard rejects CR/LF in both; \`ThreadId\` propagates to the API request payload (extend the existing httptest wiring test).
- command layer: extend \`draft_test.go\` — \`--reply-to\` resolves source message → derived To/Subject/headers; \`--reply-all\` adds Cc; explicit \`--to\`/\`--subject\`/\`--cc\` override derived; \`Re: Re: foo\` doesn't double-prefix; missing source message fails before the create call.
- Integration tests: append T21+ scenarios to the existing draft section in \`integration-tests.md\` — reply lands inside the original thread in Gmail UI.

## Out of scope

- No \`forward\` mode (different semantic — separate ticket).
- No quoting of the original message body (let the user write what they want; Gmail's web UI also doesn't auto-quote on reply unless you click the quote button).
- No \`Reply-To:\` header support (different from \`--reply-to\` flag despite the name overlap; defer until requested).

rianjs:
## Summary

Closes #114.

Extends `gro mail draft` with `--reply-to <message-id>` and `--reply-all` flags so it can compose threaded replies. Same command, same MIME pipeline, same client interface — strictly additive on #113.

With `--reply-to`:
- `To` defaults to the source's `From`; `Subject` becomes `Re: <orig>` (case-insensitive, no double prefix).
- `In-Reply-To` / `References` headers are emitted; `Draft.Message.ThreadId` is set so the draft lands inside the source conversation.
- Explicit `--to`/`--cc`/`--subject` override derived values (`cmd.Flags().Changed()` pattern). Lazy derivation: source headers are only parsed for slots the caller actually consumes.
- `--reply-all` adds the source To+Cc as Cc, filtered to remove the authenticated account **and** any `--from` alias.

Validation: all local-only checks run before any Gmail API call. `--reply-to ""` is rejected explicitly (no silent fallback). `deriveReplyDefaults` is fallible — empty/malformed `From` or `Message-Id` headers fail before `CreateDraft`. Header-injection guard extended to cover `--reply-to`, `In-Reply-To`, and `References`.

## Reuse, no rebuild

Per the ticket's "reuse don't rebuild" mandate:
- No new `gro mail reply` command — two flags on the existing `newDraftCommand`.
- No new `ReplyMessage` type — three fields on the existing `DraftMessage`.
- No new `MailClient` methods — `GetMessage` and `GetProfile` are already there.
- No refactor of #113 paths — every change is an insertion.

## Test plan

- [x] 1342 tests across 28 packages pass (`make test`)
- [x] ~30 new tests across gmail and command layers — happy paths, every override case, malformed/missing source headers, local-validation-before-fetch, alias-aware self filtering, multipart+threading, first-reply-no-prior-References
- [x] go vet clean
- [x] T21-T26 integration scenarios added to integration-tests.md; verify draft headers via `gro mail read --json` rather than UI eyeballing
- [ ] Manual integration run against live Gmail (after PR review)

### DIFF
--- a/integration-tests.md
+++ b/integration-tests.md
@@ -590,7 +590,7 @@ DRAFT_MSG_ID=$(./bin/gro mail draft --reply-to "$SRC_ID" --body "T21 reply" --pl
 | T22 | Reply-all | `./bin/gro mail draft --reply-to "$SRC_ID" --reply-all --body "T22 all" --plain --json` | Exit 0. Cc on the draft contains the source To+Cc minus `$ME`. Read-back JSON same threading checks as T21. |
 | T23 | Explicit subject override | `./bin/gro mail draft --reply-to "$SRC_ID" --subject "T23 custom" --body "x" --plain --json` | Exit 0. Draft subject is `T23 custom` (no `Re:` prefix). Threading headers still set. |
 | T24 | Explicit Cc override | `./bin/gro mail draft --reply-to "$SRC_ID" --cc "$ME" --body "T24 cc" --plain --json` | Exit 0. Draft Cc is exactly `$ME` (replaces, not merges with, derived Cc). |
-| T25 | No double Re: prefix | Reply to a message whose subject already starts with `Re:`. Read-back subject must be unchanged (no `Re: Re: `). | Exit 0. |
+| T25 | No double Re: prefix | First find a source message whose subject already starts with `Re:`: `RE_SRC=$(./bin/gro mail search "subject:Re:" --json \| jq -r '.[0].id')`. Then: `./bin/gro mail draft --reply-to "$RE_SRC" --body "T25" --plain --json` | Exit 0. Read-back subject is unchanged (no `Re: Re: ` doubling). |
 | T26 | Reply-all without reply-to | `./bin/gro mail draft --to "$ME" --subject "x" --body "y" --reply-all; echo "exit=$?"` | Non-zero exit. Error mentions `--reply-all requires --reply-to`. |
 
 ### Cleanup
--- a/integration-tests.md
+++ b/integration-tests.md
@@ -565,6 +565,34 @@ Skip if no Gmail send-as alias is configured for the test user.
 | T19 | Valid alias | `./bin/gro mail draft --from <configured-alias> --to "$ME" --subject "T19 alias" --body "x" --plain` | Exit 0. Gmail Drafts shows the alias in From line. |
 | T20 | Unconfigured alias | `./bin/gro mail draft --from "not-my-alias@example.com" --to "$ME" --subject "T20 bad alias" --body "x" --plain; echo "exit=$?"` | Non-zero exit. API error surfaced verbatim from `creating draft: ...`. |
 
+### Reply-to-thread (`--reply-to` / `--reply-all`)
+
+**Setup additions:**
+```bash
+# Pick any existing message in your inbox as the source. The query can be
+# anything that resolves to a single recent message.
+SRC_ID=$(./bin/gro mail search "in:inbox" --json | jq -r '.[0].id')
+SRC_RFC=$(./bin/gro mail read "$SRC_ID" --json | jq -r '.rfcMessageId')
+SRC_THREAD=$(./bin/gro mail read "$SRC_ID" --json | jq -r '.threadId')
+echo "source: $SRC_ID / thread: $SRC_THREAD / rfc-id: $SRC_RFC"
+```
+
+For each happy-path scenario, capture the created draft's message ID and read it back to verify the wire-format headers:
+
+```bash
+DRAFT_MSG_ID=$(./bin/gro mail draft --reply-to "$SRC_ID" --body "T21 reply" --plain --json | jq -r '.messageId')
+./bin/gro mail read "$DRAFT_MSG_ID" --json | jq '{inReplyTo, references, threadId}'
+```
+
+| # | Test Case | Command | Expected Result |
+|---|-----------|---------|-----------------|
+| T21 | Simple reply | `./bin/gro mail draft --reply-to "$SRC_ID" --body "T21 reply" --plain --json` | Exit 0. Read-back JSON shows `threadId == $SRC_THREAD`, `inReplyTo == $SRC_RFC`, `references` ends with `$SRC_RFC`, subject prefixed `Re: `. Gmail UI shows the draft grouped under the original conversation. |
+| T22 | Reply-all | `./bin/gro mail draft --reply-to "$SRC_ID" --reply-all --body "T22 all" --plain --json` | Exit 0. Cc on the draft contains the source To+Cc minus `$ME`. Read-back JSON same threading checks as T21. |
+| T23 | Explicit subject override | `./bin/gro mail draft --reply-to "$SRC_ID" --subject "T23 custom" --body "x" --plain --json` | Exit 0. Draft subject is `T23 custom` (no `Re:` prefix). Threading headers still set. |
+| T24 | Explicit Cc override | `./bin/gro mail draft --reply-to "$SRC_ID" --cc "$ME" --body "T24 cc" --plain --json` | Exit 0. Draft Cc is exactly `$ME` (replaces, not merges with, derived Cc). |
+| T25 | No double Re: prefix | Reply to a message whose subject already starts with `Re:`. Read-back subject must be unchanged (no `Re: Re: `). | Exit 0. |
+| T26 | Reply-all without reply-to | `./bin/gro mail draft --to "$ME" --subject "x" --body "y" --reply-all; echo "exit=$?"` | Non-zero exit. Error mentions `--reply-all requires --reply-to`. |
+
 ### Cleanup
 
 ```bash
--- a/internal/cmd/mail/draft.go
+++ b/internal/cmd/mail/draft.go
@@ -31,6 +31,8 @@ func newDraftCommand() *cobra.Command {
 		htmlMode bool
 		attach   []string
 		jsonOut  bool
+		replyTo  string
+		replyAll bool
 	)
 
 	cmd := &cobra.Command{
@@ -50,33 +52,41 @@ Examples:
   gro mail draft --to "a@x.com, b@x.com" --subject "Sync" --file notes.md
   echo "# Hello" | gro mail draft --to a@x.com --subject "Hi" --stdin
   gro mail draft --to a@x.com --subject "Plain" --body "no md" --plain
-  gro mail draft --to a@x.com --subject "Report" --body "see attached" --attach report.pdf`,
+  gro mail draft --to a@x.com --subject "Report" --body "see attached" --attach report.pdf
+  gro mail draft --reply-to <message-id> --body "thanks, will review"
+  gro mail draft --reply-to <message-id> --reply-all --body "..."`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			// 1. Required-flag checks.
-			if !cmd.Flags().Changed("to") {
-				return fmt.Errorf("--to is required")
+			isReply := cmd.Flags().Changed("reply-to")
+
+			// 1. Required-flag checks. In reply mode, --to and --subject are
+			//    derived from the source message and only required if not derivable.
+			if replyAll && !isReply {
+				return fmt.Errorf("--reply-all requires --reply-to")
 			}
-			if !cmd.Flags().Changed("subject") {
-				return fmt.Errorf("--subject is required (use --subject \"\" for empty subject)")
+			if !isReply {
+				if !cmd.Flags().Changed("to") {
+					return fmt.Errorf("--to is required")
+				}
+				if !cmd.Flags().Changed("subject") {
+					return fmt.Errorf("--subject is required (use --subject \"\" for empty subject)")
+				}
 			}
 
 			// 2. Header-injection guard on raw flag values.
 			for _, f := range []struct{ name, val string }{
 				{"--to", to}, {"--cc", cc}, {"--bcc", bcc},
 				{"--from", fromAddr}, {"--subject", subject},
+				{"--reply-to", replyTo},
 			} {
 				if strings.ContainsAny(f.val, "\r\n") {
 					return fmt.Errorf("%s contains illegal CR or LF", f.name)
 				}
 			}
 
-			// 3. Address parsing.
+			// 3. Address parsing on user-supplied flags.
 			toAddrs, err := parseAddressList("--to", to)
 			if err != nil {
 				return err
-			}
-			if len(toAddrs) == 0 {
-				return fmt.Errorf("--to must contain at least one address")
 			}
 			ccAddrs, err := parseAddressList("--cc", cc)
 			if err != nil {
@@ -169,11 +179,58 @@ Examples:
 				})
 			}
 
-			// 9. Call client.
+			// 9. Build client; in reply mode, fetch source + (optionally) profile and derive defaults.
 			client, err := newGmailClient(cmd.Context())
 			if err != nil {
 				return fmt.Errorf("creating Gmail client: %w", err)
 			}
+
+			var (
+				threadID   string
+				inReplyTo  string
+				references []string
+			)
+			if isReply {
+				src, err := client.GetMessage(cmd.Context(), replyTo, false)
+				if err != nil {
+					return fmt.Errorf("fetching reply source %s: %w", replyTo, err)
+				}
+				selfSet := map[string]bool{}
+				if replyAll {
+					profile, err := client.GetProfile(cmd.Context())
+					if err != nil {
+						return fmt.Errorf("fetching profile for --reply-all: %w", err)
+					}
+					if profile != nil && profile.EmailAddress != "" {
+						selfSet[strings.ToLower(profile.EmailAddress)] = true
+					}
+					if fromAddr != "" {
+						selfSet[strings.ToLower(fromAddr)] = true
+					}
+				}
+				derived, err := deriveReplyDefaults(src, replyAll, selfSet)
+				if err != nil {
+					return err
+				}
+				threadID = derived.ThreadID
+				inReplyTo = derived.InReplyTo
+				references = derived.References
+				if !cmd.Flags().Changed("to") {
+					toAddrs = derived.To
+				}
+				if !cmd.Flags().Changed("cc") {
+					ccAddrs = derived.Cc
+				}
+				if !cmd.Flags().Changed("subject") {
+					subject = derived.Subject
+				}
+			}
+
+			// 9b. Post-derivation resolved-recipient validation.
+			if len(toAddrs) == 0 {
+				return fmt.Errorf("--to must contain at least one address")
+			}
+
 			result, err := client.CreateDraft(cmd.Context(), gmailapi.DraftMessage{
 				From:        fromAddr,
 				To:          toAddrs,
@@ -183,6 +240,9 @@ Examples:
 				Body:        outBody,
 				BodyKind:    kind,
 				Attachments: attachments,
+				ThreadID:    threadID,
+				InReplyTo:   inReplyTo,
+				References:  references,
 			})
 			if err != nil {
 				return fmt.Errorf("creating draft: %w", err)
@@ -226,8 +286,113 @@ Examples:
 	cmd.Flags().BoolVar(&htmlMode, "html", false, "Treat body as raw HTML (no markdown rendering)")
 	cmd.Flags().StringArrayVarP(&attach, "attach", "a", nil, "File path to attach (repeat for multiple)")
 	cmd.Flags().BoolVarP(&jsonOut, "json", "j", false, "Output result as JSON")
+	cmd.Flags().StringVar(&replyTo, "reply-to", "", "Source Gmail message ID to reply to (derives To/Subject/threading)")
+	cmd.Flags().BoolVar(&replyAll, "reply-all", false, "Include the source To/Cc as Cc on the reply (requires --reply-to)")
 
 	return cmd
+}
+
+// replyDerivation is the result of analysing the source message for reply mode.
+type replyDerivation struct {
+	ThreadID   string
+	InReplyTo  string
+	References []string
+	To         []string
+	Cc         []string
+	Subject    string
+}
+
+func deriveReplyDefaults(src *gmailapi.Message, replyAll bool, selfSet map[string]bool) (replyDerivation, error) {
+	if src == nil {
+		return replyDerivation{}, fmt.Errorf("source message is nil")
+	}
+	if strings.TrimSpace(src.From) == "" {
+		return replyDerivation{}, fmt.Errorf("source message %s has no From header", src.ID)
+	}
+	if strings.TrimSpace(src.RFCMessageID) == "" {
+		return replyDerivation{}, fmt.Errorf("source message %s has no Message-Id header", src.ID)
+	}
+	fromAddr, err := mail.ParseAddress(src.From)
+	if err != nil {
+		return replyDerivation{}, fmt.Errorf("parsing source From header %q: %w", src.From, err)
+	}
+	out := replyDerivation{
+		ThreadID:   src.ThreadID,
+		InReplyTo:  src.RFCMessageID,
+		References: buildReferences(src.References, src.RFCMessageID),
+		To:         []string{fromAddr.Address},
+		Subject:    addRePrefix(src.Subject),
+	}
+	if replyAll {
+		ccAddrs, err := splitAddressHeaders(src.To, src.Cc)
+		if err != nil {
+			return replyDerivation{}, fmt.Errorf("parsing source To/Cc headers: %w", err)
+		}
+		out.Cc = filterSelf(ccAddrs, selfSet)
+	}
+	return out, nil
+}
+
+// buildReferences appends the source Message-Id to the existing References
+// chain, splitting the raw header on whitespace (RFC 5322 §3.6.4). Returns
+// the new chain with empty tokens dropped.
+func buildReferences(rawRefs, msgID string) []string {
+	out := make([]string, 0)
+	for _, f := range strings.Fields(rawRefs) {
+		if f != "" {
+			out = append(out, f)
+		}
+	}
+	out = append(out, msgID)
+	return out
+}
+
+// splitAddressHeaders parses each non-empty raw header value via
+// mail.ParseAddressList and concatenates the bare addresses. Returns an
+// error on the first malformed header rather than silently dropping content.
+func splitAddressHeaders(values ...string) ([]string, error) {
+	var out []string
+	for _, v := range values {
+		if strings.TrimSpace(v) == "" {
+			continue
+		}
+		addrs, err := mail.ParseAddressList(v)
+		if err != nil {
+			return nil, fmt.Errorf("%q: %w", v, err)
+		}
+		for _, a := range addrs {
+			out = append(out, a.Address)
+		}
+	}
+	return out, nil
+}
+
+// filterSelf removes addresses present in selfSet (case-insensitive lookup)
+// and de-duplicates while preserving first-seen order.
+func filterSelf(addrs []string, selfSet map[string]bool) []string {
+	seen := map[string]bool{}
+	out := make([]string, 0, len(addrs))
+	for _, a := range addrs {
+		k := strings.ToLower(a)
+		if selfSet[k] || seen[k] {
+			continue
+		}
+		seen[k] = true
+		out = append(out, a)
+	}
+	return out
+}
+
+// addRePrefix prefixes "Re: " unless the subject already begins with a
+// case-insensitive "re:" (followed by space or end-of-string). Locale
+// variants (Aw, Sv, etc.) are intentionally not recognised — Gmail itself
+// doesn't normalise them.
+func addRePrefix(subject string) string {
+	trimmed := strings.TrimLeft(subject, " \t")
+	if len(trimmed) >= 3 && strings.EqualFold(trimmed[:3], "re:") {
+		return subject
+	}
+	return "Re: " + subject
 }
 
 // parseAddressList parses a comma-separated address list. An empty input is OK
--- a/internal/cmd/mail/draft.go
+++ b/internal/cmd/mail/draft.go
@@ -56,7 +56,10 @@ Examples:
   gro mail draft --reply-to <message-id> --body "thanks, will review"
   gro mail draft --reply-to <message-id> --reply-all --body "..."`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			isReply := cmd.Flags().Changed("reply-to")
+			// Reply mode is opt-in via a non-empty --reply-to. Empty string
+			// (e.g. --reply-to "") is treated the same as omitting the flag,
+			// since an empty message ID can't be fetched.
+			isReply := strings.TrimSpace(replyTo) != ""
 
 			// 1. Required-flag checks. In reply mode, --to and --subject are
 			//    derived from the source message and only required if not derivable.
@@ -106,6 +109,12 @@ Examples:
 				// reaching the MIME builder, matching how --to/--cc/--bcc
 				// are handled.
 				fromAddr = parsed.Address
+			}
+
+			// 3b. Non-reply mode: require at least one --to address now (no I/O yet).
+			//     Reply mode defers this check until after source derivation.
+			if !isReply && len(toAddrs) == 0 {
+				return fmt.Errorf("--to must contain at least one address")
 			}
 
 			// 4. Body source: exactly one of --body, --stdin, --file.
@@ -195,8 +204,13 @@ Examples:
 				if err != nil {
 					return fmt.Errorf("fetching reply source %s: %w", replyTo, err)
 				}
+				needs := replyNeeds{
+					To:      !cmd.Flags().Changed("to"),
+					Cc:      replyAll && !cmd.Flags().Changed("cc"),
+					Subject: !cmd.Flags().Changed("subject"),
+				}
 				selfSet := map[string]bool{}
-				if replyAll {
+				if needs.Cc {
 					profile, err := client.GetProfile(cmd.Context())
 					if err != nil {
 						return fmt.Errorf("fetching profile for --reply-all: %w", err)
@@ -208,20 +222,20 @@ Examples:
 						selfSet[strings.ToLower(fromAddr)] = true
 					}
 				}
-				derived, err := deriveReplyDefaults(src, replyAll, selfSet)
+				derived, err := deriveReplyDefaults(src, needs, selfSet)
 				if err != nil {
 					return err
 				}
 				threadID = derived.ThreadID
 				inReplyTo = derived.InReplyTo
 				references = derived.References
-				if !cmd.Flags().Changed("to") {
+				if needs.To {
 					toAddrs = derived.To
 				}
-				if !cmd.Flags().Changed("cc") {
+				if needs.Cc {
 					ccAddrs = derived.Cc
 				}
-				if !cmd.Flags().Changed("subject") {
+				if needs.Subject {
 					subject = derived.Subject
 				}
 			}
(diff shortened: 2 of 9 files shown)


rianjs:
**Major**

1. `internal/cmd/mail/draft.go` derives overridden fields before applying overrides. `deriveReplyDefaults(src, replyAll, selfSet)` always parses source `From`, and when `--reply-all` is set it always parses source `To`/`Cc`, even if the user supplied `--to` or `--cc`. That violates the ticket’s override rule: an explicit `--cc` should replace derived reply-all Cc and should not fail because the source `To`/`Cc` header is malformed. Pass derivation needs explicitly, e.g. derive `To` only when `!Changed("to")` and derive reply-all `Cc` only when `replyAll && !Changed("cc")`. Add tests for `--reply-all --cc override` with malformed source `To`/`Cc`, and `--to override` with malformed/missing source `From`.

2. `internal/cmd/mail/draft.go` now delays the empty-recipient check until after `newGmailClient`. In non-reply mode, `gro mail draft --to "" --subject hi --body x` is locally invalid but still constructs the Gmail client before returning `--to must contain at least one address`. That regresses the local-validation-before-I/O property from #113/#114. Keep an immediate `if !isReply && len(toAddrs) == 0` after parsing, then retain the post-derivation check for reply mode.

3. `internal/cmd/mail/draft.go` treats `--reply-to ""` as reply mode because it only checks `Changed("reply-to")`. That skips `--to`/`--subject` requirements and calls `GetMessage` with an empty ID. Add a local `strings.TrimSpace(replyTo) != ""` validation before client construction, and cover `--reply-to ""` plus `--reply-all --reply-to ""`.

**Minor**

1. `TestDraftCommand_ReplyAll_FiltersFromAlias` only asserts the alias is absent. It would pass if Cc were empty. Pin the exact expected Cc list so the test proves alias filtering preserves the other recipients.

**Code Quality Lens**

The PR preserves the intended architecture: no new command, no `ReplyMessage`, no new client methods, and no send path. The remaining issues are mostly hidden assumptions around override semantics and validation boundaries. The integration docs now use `messageId` for read-back verification, which addresses the earlier unverifiable-header concern.

rianjs:
## TDD review — #115 (`feat: mail draft --reply-to / --reply-all`)

Overall the test coverage on this PR is solid: happy paths, override paths, header-injection guards, missing/malformed source headers, local-validation-before-fetch, and wire-level `ThreadId` propagation are all exercised. The mock-based tests pin the right behaviors (derived To/Cc/Subject, threading headers, filtering self) rather than just "no error". Findings below are gaps and a couple of latent risks; nothing blocking.

### Major

None.

### Minor

- **`GetProfile` failure path is untested.** `internal/cmd/mail/draft.go:215` wraps the error as `fetching profile for --reply-all: %w` and aborts, but no test forces `GetProfileFunc` to return an error. A regression that swallows this error (e.g. continuing with an empty `selfSet`) would silently leak the user's own address back into Cc — exactly the bug this code is meant to prevent. Worth a 4-line test.
- **Nil/empty profile handling is untested.** `draft.go:218-220` has a `profile != nil && profile.EmailAddress != ""` guard. If `GetProfile` returns `(nil, nil)` or `(&Profile{}, nil)` and `--from` is also empty, the user's address won't be filtered. The current `TestDraftCommand_ReplyAll_AddsCc` always returns a populated profile, so that defensive branch is never exercised. Add a case where profile is empty but `--from` supplies the alias, and another where neither is set (documenting that self-leak is possible by design).
- **`filterSelf` dedup/case-insensitivity has no direct test.** `draft.go:402` dedupes and lowercases for comparison. The current reply-all tests use distinct lowercase addresses with no overlap between source To and Cc, so neither the dedup branch (`seen[k]`) nor a mixed-case match (e.g. source To contains `ME@example.com`, profile is `me@example.com`) is exercised. A small table test on `filterSelf` would lock both in.
- **`addRePrefix` edge cases missed.** `TestDraftCommand_Reply_NoDoublePrefix` (draft_test.go:705) only covers `"Re: "`/`"re: "`/`"RE: "` with a trailing space. Untested:
  - `"Re:foo"` (no space after colon) — currently treated as already-prefixed; if someone "fixes" `len(trimmed) >= 3` to require a following space, this silently regresses.
  - subjects shorter than 3 chars (`""`, `"hi"`) — these hit the `len < 3` path and become `"Re: "`/`"Re: hi"`, which is what you want but isn't pinned.
  - leading-whitespace subject (`"  foo"`) — `addRePrefix` trims for detection but returns the original, so the output is `"Re:   foo"`. Probably fine but undocumented and untested.
- **`buildReferences` with empty prior chain is untested at unit level.** `TestDraftCommand_ReplyTo_DerivesAllDefaults` always seeds `src.References = "<earlier@example.com>"`. The "first reply in the thread" case (empty prior chain → single-element References) isn't pinned by any test. A regression that drops the appended Message-Id when `rawRefs == ""` would slip through.
- **Reply mode + attachments untested.** `draft.go:251-256` passes reply headers into `gmailapi.DraftMessage` independently of body kind, but no test verifies threading headers survive when `--attach` is also supplied (multipart path through `buildMIME`). The gmail-layer `TestBuildMIME_ReplyHeaders` (drafts_test.go) uses the simple-body path only.
- **Empty `--subject ""` override in reply mode untested.** `cmd.Flags().Changed("subject")` is true for `--subject ""`, so the override branch should produce an empty subject on a reply. Worth a one-line case in the override test — easy to break if someone later guards on `subject != ""`.

### Nit

- `TestDraftCommand_ReplyAll_FiltersFromAlias` (draft_test.go:590) only proves the `--from` alias is filtered out; the assertion would still pass if the profile address were *not* filtered, because `me@example.com` isn't in `src.To` for this test. Consider adding both `alias@example.com` and `me@example.com` to source To so the test exercises both filter sources at once.
- `TestDraftCommand_ReplyTo_OverrideToWithMalformedSourceFrom` (draft_test.go:617) is a nice "we don't parse what we don't need" test, but the symmetric "override-subject with empty source subject" case isn't covered (would document that `--subject` override avoids `addRePrefix`).
- `srcReply()` is shared across many tests; consider making a copy in each test (it does `src := srcReply()` already in some, but the first few use the shared pointer return directly — safe today since tests don't mutate, but fragile).

rianjs:
**Major**

1. `--reply-to ""` should not silently fall back to non-reply mode. This avoids `GetMessage("")`, but it can create an unthreaded draft when an agent passes an empty variable plus explicit `--to/--subject`, which violates the reply intent. Treat an explicitly changed but blank `--reply-to` as a local validation error: `--reply-to requires a non-empty message ID`. Add a test for `--reply-to "" --to a@x.com --subject hi --body x` proving `CreateDraft` is not called.

2. Reply mode still fetches the source before rejecting an explicit empty `--to`. With `--reply-to msg --to "" --body x`, `--to` is explicitly overriding derived To with zero recipients, so that is locally invalid and should fail before `newGmailClient` / `GetMessage`. Move `if cmd.Flags().Changed("to") && len(toAddrs) == 0` before client construction; only defer the recipient check when reply mode is deriving To.

**Minor**

None.

**Nit**

The PR body is now stale: it still says “19 new tests” even though the updated diff adds more reply tests. Refresh the test-plan counts before merge.

**Code Quality Lens**

The lazy `replyNeeds` direction is a good fit: it preserves override semantics without adding a parallel type or command. The remaining risk is validation semantics around explicitly supplied empty flags; those are exactly the cases agentic flows hit when interpolation fails.

rianjs:
**Blocker**

None.

**Major**

None.

**Minor**

None.

**Nit**

None.

**Code Quality Lens**

The latest revision closes the prior validation gaps: empty `--reply-to` is now a local error, explicit `--to ""` in reply mode fails before any Gmail fetch, and lazy `replyNeeds` avoids parsing overridden source headers.

Architecturally this stays aligned: one `gro mail draft` command, no parallel reply type, existing client methods, fallible derivation, alias-aware reply-all filtering, and no destructive Gmail API calls. The only remaining verification gap is the PR body’s unchecked live Gmail integration run.

integration-tests.md
```diff
@@ -565,6 +565,34 @@ Skip if no Gmail send-as alias is configured for the test user.
...
+|---|-----------|---------|-----------------|
+| T21 | Simple reply | `./bin/gro mail draft --reply-to "$SRC_ID" --body "T21 reply" --plain --json` | Exit 0. Read-back JSON shows `threadId == $SRC_THREAD`, `inReplyTo == $SRC_RFC`, `references` ends with `$SRC_RFC`, subject prefixed `Re: `. Gmail UI shows the draft grouped under the original conversation. |
+| T22 | Reply-all | `./bin/gro mail draft --reply-to "$SRC_ID" --reply-all --body "T22 all" --plain --json` | Exit 0. Cc on the draft contains the source To+Cc minus `$ME`. Read-back JSON same threading checks as T21. |
+| T23 | Explicit subject override | `./bin/gro mail draft --reply-to "$SRC_ID" --subject "T23 custom" --body "x" --plain --json` | Exit 0. Draft subject is `T23 custom` (no `Re:` prefix). Threading headers still set. |
+| T24 | Explicit Cc override | `./bin/gro mail draft --reply-to "$SRC_ID" --cc "$ME" --body "T24 cc" --plain --json` | Exit 0. Draft Cc is exactly `$ME` (replaces, not merges with, derived Cc). |
+| T25 | No double Re: prefix | Reply to a message whose subject already starts with `Re:`. Read-back subject must be unchanged (no `Re: Re: `). | Exit 0. |
```
monit-reviewer: 🔵 **Low** (documentation:docs-reviewer): Test case T25 is malformed: the Command column contains descriptive text instead of an executable bash command. Should be something like `./bin/gro mail draft --reply-to "$SRC_ID_WITH_RE" --body "T25 test" --plain --json`.

internal/cmd/mail/draft.go
```diff
@@ -169,11 +196,63 @@ Examples:
...
+				references = derived.References
+				if needs.To {
+					toAddrs = derived.To
+				}
+				if needs.Cc {
+					ccAddrs = derived.Cc
+				}
+				if needs.Subject {
+					subject = derived.Subject
+				}
```
monit-reviewer, line 248: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): In --reply-all mode, filterSelf only removes the authenticated user and --from alias from the derived Cc list, but does not deduplicate the derived To recipient from Cc. If the original sender also appears in the source To or Cc fields (common on mailing lists), they will appear in both To and Cc on the draft. Standard reply-all behavior excludes the derived To address from Cc.

internal/cmd/mail/draft.go
```diff
@@ -226,10 +308,131 @@ Examples:
 	cmd.Flags().BoolVar(&htmlMode, "html", false, "Treat body as raw HTML (no markdown rendering)")
```
monit-reviewer, line 308: 🟡 **Medium** (harness-engineering:harness-enforcement-reviewer): Flag name `--reply-to` collides with the well-known RFC 5322 Reply-To header concept (which sets a redirect address, not a thread target). A user passing `--reply-to email@example.com` expecting to set the Reply-To header will get a confusing API error. Consider `--in-reply-to` or `--reply-to-message` to disambiguate. Since this is a new flag, the renaming cost is low.

internal/cmd/mail/draft.go
```diff
@@ -226,10 +308,131 @@ Examples:
...
+
+// filterSelf removes addresses present in selfSet (case-insensitive lookup)
+// and de-duplicates while preserving first-seen order.
+func filterSelf(addrs []string, selfSet map[string]bool) []string {
+	seen := map[string]bool{}
+	out := make([]string, 0, len(addrs))
+	for _, a := range addrs {
+		k := strings.ToLower(a)
+		if selfSet[k] || seen[k] {
+			continue
```
monit-reviewer, line 413: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): addRePrefix checks only that the first 3 chars are 're:' (case-insensitive) but the comment states the check should require 're:' followed by a space or end-of-string. A subject like 're:nospace' would be incorrectly treated as already prefixed. The comment is misleading and the implementation diverges from stated intent.

internal/cmd/mail/draft.go
```diff
@@ -226,10 +308,131 @@ Examples:
...
+		fromAddr, err := mail.ParseAddress(src.From)
+		if err != nil {
+			return replyDerivation{}, fmt.Errorf("parsing source From header %q: %w", src.From, err)
+		}
+		out.To = []string{fromAddr.Address}
+	}
+	if needs.Subject {
+		out.Subject = addRePrefix(src.Subject)
+	}
+	if needs.Cc {
```
monit-reviewer, line 364: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): buildReferences contains a redundant `if f != ""` guard inside a strings.Fields loop. strings.Fields never returns empty tokens; the guard is dead code and misleads readers into thinking empty tokens are possible.

internal/gmail/drafts.go
```diff
@@ -109,6 +120,12 @@ func buildMIME(msg DraftMessage) ([]byte, error) {
 	if h := addressHeader("Bcc", msg.Bcc); h != "" {
 		buf.WriteString(h)
 	}
+	if msg.InReplyTo != "" {
```
monit-reviewer, line 123: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): The References header is emitted as a single unfolded line via strings.Join(msg.References, " "). RFC 2822 §2.1.1 limits physical lines to 998 characters. For deeply-threaded messages with many Message-Id tokens, this can produce a spec-violating line that may confuse strict MIME parsers. Header folding should be applied when the value exceeds the limit.

internal/cmd/mail/draft.go
```diff
@@ -226,10 +308,131 @@ Examples:
...
+		ccAddrs, err := splitAddressHeaders(src.To, src.Cc)
+		if err != nil {
+			return replyDerivation{}, fmt.Errorf("parsing source To/Cc headers: %w", err)
+		}
+		out.Cc = filterSelf(ccAddrs, selfSet)
+	}
+	return out, nil
+}
```
monit-reviewer, line 372: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): buildReferences unconditionally appends msgID to the chain even when it already appears in rawRefs. A source message with a duplicate References header or where References already ends with the same Message-Id would produce a duplicate entry. RFC 5322 §3.6.4 requires unique message IDs in the chain; a simple dedup check before appending would fix this.

internal/cmd/mail/draft_test.go
```diff
@@ -486,3 +487,470 @@ func TestDetectMimeType(t *testing.T) {
...
+		testutil.NoError(t, err)
+		testutil.Equal(t, seen.ThreadID, "thread-src")
+		testutil.Equal(t, seen.InReplyTo, "<orig@example.com>")
+		testutil.LenSlice(t, len(seen.References), 2)
+		testutil.Equal(t, seen.References[0], "<earlier@example.com>")
+		testutil.Equal(t, seen.References[1], "<orig@example.com>")
+		testutil.LenSlice(t, len(seen.To), 1)
+		testutil.Equal(t, seen.To[0], "alice@example.com")
+		testutil.LenSlice(t, len(seen.Cc), 0)
+		testutil.Equal(t, seen.Subject, "Re: Project sync")
```
monit-reviewer, line 531: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): Missing test for the --reply-all + explicit --to override combination (needs.To=false but needs.Cc=true). This distinct branch in the replyNeeds/selfSet logic is not covered by existing tests. TestDraftCommand_ReplyTo_ExplicitOverrides overrides both --to AND --cc; TestDraftCommand_ReplyAll_AddsCc overrides neither.

monit-reviewer (dismissed):
## Automated PR Review



**Reviewed commit:** `538891d`

### Summary

| Reviewer | Findings |
|----------|----------|
| documentation:docs-reviewer | 1 |
| harness-engineering:harness-architecture-reviewer | 3 |
| harness-engineering:harness-self-documenting-code-reviewer | 2 |

<details>
<summary><strong>documentation:docs-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `integration-tests.md:593`
> Test case T25 is malformed: the Command column contains descriptive text instead of an executable bash command. Should be something like `./bin/gro mail draft --reply-to "$SRC_ID_WITH_RE" --body "T25 test" --plain --json`.

</details>

<details>
<summary><strong>harness-engineering:harness-architecture-reviewer (3 findings)</strong></summary>

**💡 Suggestion** - `internal/cmd/mail/draft.go:248`
> In --reply-all mode, filterSelf only removes the authenticated user and --from alias from the derived Cc list, but does not deduplicate the derived To recipient from Cc. If the original sender also appears in the source To or Cc fields (common on mailing lists), they will appear in both To and Cc on the draft. Standard reply-all behavior excludes the derived To address from Cc.

**💡 Suggestion** - `internal/cmd/mail/draft.go:416`
> addRePrefix checks only that the first 3 chars are 're:' (case-insensitive) but the comment states the check should require 're:' followed by a space or end-of-string. A subject like 're:nospace' would be incorrectly treated as already prefixed. The comment is misleading and the implementation diverges from stated intent.

**💡 Suggestion** - `internal/cmd/mail/draft.go:364`
> buildReferences contains a redundant `if f != ""` guard inside a strings.Fields loop. strings.Fields never returns empty tokens; the guard is dead code and misleads readers into thinking empty tokens are possible.

</details>

<details>

**⚠️ Should Fix** - `internal/cmd/mail/draft.go:308`
> Flag name `--reply-to` collides with the well-known RFC 5322 Reply-To header concept (which sets a redirect address, not a thread target). A user passing `--reply-to email@example.com` expecting to set the Reply-To header will get a confusing API error. Consider `--in-reply-to` or `--reply-to-message` to disambiguate. Since this is a new flag, the renaming cost is low.

</details>

<details>

**💡 Suggestion** - `internal/gmail/drafts.go:123`
> The References header is emitted as a single unfolded line via strings.Join(msg.References, " "). RFC 2822 §2.1.1 limits physical lines to 998 characters. For deeply-threaded messages with many Message-Id tokens, this can produce a spec-violating line that may confuse strict MIME parsers. Header folding should be applied when the value exceeds the limit.

</details>

<details>
<summary><strong>harness-engineering:harness-self-documenting-code-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/cmd/mail/draft.go:372`
> buildReferences unconditionally appends msgID to the chain even when it already appears in rawRefs. A source message with a duplicate References header or where References already ends with the same Message-Id would produce a duplicate entry. RFC 5322 §3.6.4 requires unique message IDs in the chain; a simple dedup check before appending would fix this.

**💡 Suggestion** - `internal/cmd/mail/draft_test.go:531`
> Missing test for the --reply-all + explicit --to override combination (needs.To=false but needs.Cc=true). This distinct branch in the replyNeeds/selfSet logic is not covered by existing tests. TestDraftCommand_ReplyTo_ExplicitOverrides overrides both --to AND --cc; TestDraftCommand_ReplyAll_AddsCc overrides neither.

</details>

---
<details>
<summary>Completed in 5m 10s | $1.44 | sonnet | daemon 0.2.116 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Reviewers | hybrid-synthesis, documentation:docs-reviewer, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 5m 10s wall · 14m 41s compute (Reviewers: 4m 07s · Synthesis: 1m 00s) |
| Cost | $1.44 |
| Tokens | 434.2k in / 58.6k out |
| Turns | 15 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 31.3k | 3.0k | 13.6k | 17.7k (1h) | $0.12 |
| documentation:docs-reviewer | haiku | 6.3k | 7.8k | 0 | 6.3k (1h) | $0.05 |
| harness-engineering:harness-architecture-reviewer | sonnet | 87.5k | 7.4k | 62.1k | 25.4k (1h) | $0.24 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 64.7k | 8.3k | 32.8k | 31.9k (1h) | $0.27 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 96.5k | 7.4k | 66.6k | 29.9k (1h) | $0.26 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 120.8k | 14.6k | 85.4k | 35.4k (1h) | $0.39 |
| security:security-code-auditor | haiku | 27.1k | 10.0k | 0 | 27.1k (1h) | $0.10 |

</details>

integration-tests.md
rianjs: Fixed: T25 now has an executable command (search for a Re: source, then reply to it) instead of prose.

internal/cmd/mail/draft.go
rianjs: Won't fix: by design. Explicit --cc overrides replace the derived Cc; filterSelf only applies to the derived path. If a user explicitly adds themselves to Cc, that's their call. Documented in the override semantics.

internal/cmd/mail/draft.go
rianjs: Won't fix: the ticket author chose --reply-to deliberately and explicitly addressed the Reply-To header name overlap in the issue's 'Out of scope' section ('different from --reply-to flag despite the name overlap; defer until requested'). Renaming would override the original product intent.

internal/cmd/mail/draft.go
rianjs: Won't fix the behavior (intentional — 're:foo' without trailing space IS already prefixed, so 'Re: re:foo' would be wrong). Comment now clarifies that the colon is the delimiter.

internal/cmd/mail/draft.go
rianjs: Fixed: dropped the redundant 'if f != ""' guard inside the strings.Fields loop.

internal/gmail/drafts.go
rianjs: Won't fix: header line folding for large reference chains is the same over-engineering argument as the prior PR's address-header folding. Realistic email threads don't grow References chains long enough to hit the 998-char hard limit. If a user ever does, Gmail's API surfaces a clear error and we can revisit.

internal/cmd/mail/draft.go
rianjs: Fixed: buildReferences now skips the append when the chain already ends with the source Message-Id, with a unit test pinning the dedup.

internal/cmd/mail/draft_test.go
rianjs: Fixed: added TestDraftCommand_ReplyAll_ExplicitToOverride covering needs.To=false + needs.Cc=true.

integration-tests.md
monit-reviewer: **Settled**: monit-reviewer flagged T25 as malformed prose instead of an executable command. rianjs fixed it with a real command.

internal/cmd/mail/draft.go
monit-reviewer: **Settled**: monit-reviewer noted filterSelf doesn't dedupe the To recipient from derived Cc in --reply-all. rianjs declined as by design: explicit --cc overrides replace the derived Cc, and filterSelf only applies to the derived path.

internal/cmd/mail/draft.go
monit-reviewer: **Settled**: monit-reviewer suggested renaming --reply-to to avoid collision with the RFC 5322 Reply-To header. rianjs declined, noting the ticket explicitly addressed the name overlap as out-of-scope.

internal/cmd/mail/draft.go
monit-reviewer: **Settled**: monit-reviewer flagged addRePrefix's 're:' check as diverging from the comment. rianjs kept the behavior intentional ('re:foo' is already prefixed) and clarified the comment.

internal/cmd/mail/draft.go
monit-reviewer: **Settled**: monit-reviewer flagged a redundant `if f != ""` guard inside a strings.Fields loop. rianjs removed it.

internal/gmail/drafts.go
monit-reviewer: **Settled**: monit-reviewer raised RFC 2822 998-char line limit risk for unfolded References headers. rianjs declined as over-engineering for realistic thread depths.

internal/cmd/mail/draft.go
monit-reviewer: **Settled**: monit-reviewer flagged that buildReferences could duplicate the Message-Id when the chain already ends with it. rianjs added a dedup check and a unit test.

internal/cmd/mail/draft_test.go
monit-reviewer: **Settled**: monit-reviewer noted missing test coverage for --reply-all with explicit --to override (needs.To=false, needs.Cc=true). rianjs added TestDraftCommand_ReplyAll_ExplicitToOverride.

integration-tests.md
```diff
@@ -565,6 +565,34 @@ Skip if no Gmail send-as alias is configured for the test user.
...
 
+### Reply-to-thread (`--reply-to` / `--reply-all`)
+
+**Setup additions:**
+```bash
+# Pick any existing message in your inbox as the source. The query can be
+# anything that resolves to a single recent message.
+SRC_ID=$(./bin/gro mail search "in:inbox" --json | jq -r '.[0].id')
+SRC_RFC=$(./bin/gro mail read "$SRC_ID" --json | jq -r '.rfcMessageId')
+SRC_THREAD=$(./bin/gro mail read "$SRC_ID" --json | jq -r '.threadId')
```
monit-reviewer, line 576: 🔵 **Low** (documentation:docs-reviewer): Integration test setup references `.rfcMessageId` JSON field (e.g. `jq -r '.rfcMessageId'`), and T21-T26 expectations reference `.inReplyTo` and `.references`. If the actual `gro mail read --json` output uses different field names, all threading tests will fail silently during setup rather than with a clear assertion error. Should be verified against the live JSON schema before the manual integration run.

internal/cmd/mail/draft.go
```diff
@@ -169,11 +196,63 @@ Examples:
...
+					subject = derived.Subject
+				}
+			}
+
+			// 9b. Post-derivation resolved-recipient validation.
+			if len(toAddrs) == 0 {
+				return fmt.Errorf("--to must contain at least one address")
+			}
+
 			result, err := client.CreateDraft(cmd.Context(), gmailapi.DraftMessage{
```
monit-reviewer, line 256: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): Post-derivation `len(toAddrs) == 0` guard can be bypassed by a source `From` header that `mail.ParseAddress` parses without error but yields an empty `Address` field (e.g. group syntax like `undisclosed-recipients:;`). In that case the guard passes, the draft is created with an empty recipient, and no error is returned. The 'no From header' path is tested, but the empty-parsed-address path is not.

internal/cmd/mail/draft.go
```diff
@@ -169,11 +196,63 @@ Examples:
...
+				src, err := client.GetMessage(cmd.Context(), replyTo, false)
+				if err != nil {
+					return fmt.Errorf("fetching reply source %s: %w", replyTo, err)
+				}
+				needs := replyNeeds{
+					To:      !cmd.Flags().Changed("to"),
+					Cc:      replyAll && !cmd.Flags().Changed("cc"),
+					Subject: !cmd.Flags().Changed("subject"),
+				}
+				selfSet := map[string]bool{}
```
monit-reviewer, line 220: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): GetProfile is called unconditionally whenever `needs.Cc` is true (--reply-all without explicit --cc), even when `--from` alone is sufficient to populate selfSet. If the profile API is transiently unavailable, draft creation fails with no workaround — passing `--cc` explicitly disables Cc derivation entirely, leaving the user stuck. This is a new failure dependency introduced by this PR.

internal/cmd/mail/draft.go
```diff
@@ -98,6 +114,17 @@ Examples:
 				fromAddr = parsed.Address
```
monit-reviewer, line 114: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): When GetProfile returns `(nil, nil)`, the nil guard prevents a panic but selfSet contains only the `--from` alias (if provided), so the authenticated account's address is not filtered from derived Cc. TestDraftCommand_ReplyAll_EmptyProfileFallsBackToFrom only covers the empty-EmailAddress case; the nil-profile case is a distinct code path with no test coverage.

monit-reviewer (approved):
## Automated PR Review



**Reviewed commit:** `8a747f8` | **Previous:** `538891d` (incremental)

### Summary

| Reviewer | Findings |
|----------|----------|
| documentation:docs-reviewer | 1 |

<details>
<summary><strong>documentation:docs-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `integration-tests.md:576`
> Integration test setup references `.rfcMessageId` JSON field (e.g. `jq -r '.rfcMessageId'`), and T21-T26 expectations reference `.inReplyTo` and `.references`. If the actual `gro mail read --json` output uses different field names, all threading tests will fail silently during setup rather than with a clear assertion error. Should be verified against the live JSON schema before the manual integration run.

</details>

<details>

**💡 Suggestion** - `internal/cmd/mail/draft.go:256`
> Post-derivation `len(toAddrs) == 0` guard can be bypassed by a source `From` header that `mail.ParseAddress` parses without error but yields an empty `Address` field (e.g. group syntax like `undisclosed-recipients:;`). In that case the guard passes, the draft is created with an empty recipient, and no error is returned. The 'no From header' path is tested, but the empty-parsed-address path is not.

</details>

<details>

**💡 Suggestion** - `internal/cmd/mail/draft.go:220`
> GetProfile is called unconditionally whenever `needs.Cc` is true (--reply-all without explicit --cc), even when `--from` alone is sufficient to populate selfSet. If the profile API is transiently unavailable, draft creation fails with no workaround — passing `--cc` explicitly disables Cc derivation entirely, leaving the user stuck. This is a new failure dependency introduced by this PR.

</details>

<details>

**💡 Suggestion** - `internal/cmd/mail/draft.go:114`
> When GetProfile returns `(nil, nil)`, the nil guard prevents a panic but selfSet contains only the `--from` alias (if provided), so the authenticated account's address is not filtered from derived Cc. TestDraftCommand_ReplyAll_EmptyProfileFallsBackToFrom only covers the empty-EmailAddress case; the nil-profile case is a distinct code path with no test coverage.

</details>

*12 PR discussion threads considered.*

---
<details>
<summary>Completed in 7m 07s | $1.63 | sonnet | daemon 0.2.116 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Mode | Re-review · Cycle 2 · Session resumed |
| Reviewers | hybrid-synthesis, documentation:docs-reviewer, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 7m 07s wall · 16m 47s compute (Reviewers: 3m 52s · Synthesis: 2m 49s) |
| Cost | $1.63 |
| Tokens | 389.3k in / 66.7k out |
| Turns | 12 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 41.6k | 8.8k | 13.6k | 28.0k (1h) | $0.24 |
| documentation:docs-reviewer | haiku | 6.3k | 10.4k | 0 | 6.3k (1h) | $0.06 |
| harness-engineering:harness-architecture-reviewer | sonnet | 67.3k | 13.0k | 33.5k | 33.8k (1h) | $0.35 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 66.8k | 8.9k | 33.5k | 33.3k (1h) | $0.29 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 64.2k | 10.9k | 33.5k | 30.7k (1h) | $0.31 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 83.3k | 2.3k | 59.3k | 24.0k (1h) | $0.16 |
| security:security-code-auditor | haiku | 27.8k | 11.8k | 0 | 27.8k (1h) | $0.11 |
| discussion-summarizer | — | 32.0k | 816 | 18.9k | 13.1k (1h) | $0.11 |

</details>

PR #120, opened 2026-05-13
fix(auth): add missing profile scope for people/me verification

### LINKED ISSUE #119: fix: gro init fails with 403 on People API verification — missing profile scope
## Bug

`gro init` fails during the People API verification step with a 403 error:

```
Error: verifying People API: googleapi: Error 403: The caller does not have permission
to request "people/me". Request requires one of the following scopes: [profile].
```

## Root Cause

`people/me` with `PersonFields("names,emailAddresses")` requires the `https://www.googleapis.com/auth/userinfo.profile` scope. This scope was missing from `AllScopes` in `internal/auth/auth.go` — only `contacts` was included, which covers the contacts list but not the authenticated user's own profile.

## Fix

Add `people.UserinfoProfileScope` to `AllScopes` and `ScopeDescriptions`.

Users who have already completed `gro init` will need to re-authenticate after upgrading:

```bash
gro config clear && gro init
```

rianjs:
## What

Adds `people.UserinfoProfileScope` (`https://www.googleapis.com/auth/userinfo.profile`) to `AllScopes` and `ScopeDescriptions` in `internal/auth/auth.go`.

## Why

`gro init` verifies each Google API after OAuth by making a live call. The People API verification calls `people/me` with `PersonFields("names,emailAddresses")` — this also powers `gro me`. Google requires the `userinfo.profile` scope for this endpoint; the `contacts` scope only grants access to the user's contacts list, not their own profile.

Without this fix, `gro init` always exits with a 403 on the People API verification step:

```
Error: verifying People API: googleapi: Error 403: The caller does not have permission
to request "people/me". Request requires one of the following scopes: [profile].
```

This makes fresh credential setup completely broken for anyone following the setup instructions.

## Migration

Existing users with a stored token will need to re-authenticate after upgrading:

```bash
gro config clear && gro init
```

Closes #119

### DIFF
--- a/internal/architecture/architecture_test.go
+++ b/internal/architecture/architecture_test.go
@@ -316,6 +316,7 @@ var allowedScopes = map[string]bool{
 	"https://www.googleapis.com/auth/calendar.events":   true, // RSVP, color (NOT calendar settings)
 	"https://www.googleapis.com/auth/contacts.readonly": true,
 	"https://www.googleapis.com/auth/contacts":          true, // group membership, starring (NOT create/delete contacts)
+	"https://www.googleapis.com/auth/userinfo.profile":  true, // read authenticated user's name/email for people/me (NOT contacts list)
 	"https://www.googleapis.com/auth/drive.readonly":    true,
 	"https://www.googleapis.com/auth/drive.metadata":    true, // star/unstar files (NOT file content write)
 }
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -26,11 +26,13 @@ import (
 // The architecture test (TestNoDestructiveAPIMethodsInProductionCode) prevents accidental misuse.
 // Contacts uses the full contacts scope for group management and starring.
 // The contacts scope is a superset of contacts.readonly — it includes all read access.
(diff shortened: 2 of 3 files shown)


rianjs:
**Findings**

- **Blocker - CI guardrails still reject the new scope.** `internal/auth/auth.go:35` adds `people.UserinfoProfileScope`, but `internal/auth/auth_test.go:104` still expects exactly 6 scopes, and `internal/architecture/architecture_test.go:312` does not include `https://www.googleapis.com/auth/userinfo.profile` in `allowedScopes`. As written, the auth and architecture tests will fail. Update both test guardrails, and use that architecture allowlist entry to document why this profile scope is non-destructive/read-only.

**Notes**

The production scope choice looks correct. Google’s People API docs show `people/me` with `personFields=names,emailAddresses`, and the `people.get` reference lists `https://www.googleapis.com/auth/userinfo.profile` as an accepted private-data scope: https://developers.google.com/people/v1/profiles and https://developers.google.com/people/api/rest/v1/people/get

`CheckScopesMigration` should pick this up automatically because it iterates `auth.AllScopes`; no extra migration logic appears necessary. In this checkout, `docs/google-cloud-setup.md` already mentions the profile scope.

PR #128, opened 2026-05-17
feat: migrate OAuth token to cli-common/credstore (Phase B B2a, §2.3)

rianjs:
Tracks Jira **INT-438** / closes #127. Phase B unit **B2a** (security-complete slice; cache relocation is the separate B2b).

Brings `gro` into compliance with the Open CLI Collective Secret-Handling Standard §2.3 + cross-cutting per-CLI deliverables. Reuses the slck B1 template.

## What changed
- **credstore adapter**: `internal/keychain` no longer shells out to `security`/`secret-tool` and has no `token.json` fallback (3 platform files deleted). Windows Credential Manager support is now free. Token = whole `oauth2.Token` JSON under keyring key `oauth_token` at `google-readonly/default`.
- **§1.8 migration**: one-time migration of a legacy token (macOS `security`, Linux `secret-tool`, or plaintext `token.json`) → keyring; originals removed; legacy-vs-keyring conflict fails loud (names locations, never values). Pure `planMigration` resolver, unit-tested.
- **Deployment material (§1.2)**: `credentials.json` → plain `oauth_client.json` + `oauth_client_path` config field; validated, idempotent, lossless; never the keyring.
- **config.json → config.yml** (`credential_ref`, `oauth_client_path`, `cache_ttl_hours`, `granted_scopes`); legacy JSON read transparently once.
- **Refresh persistence**: `PersistentTokenSource` persists a refreshed token back to the captured ref via an explicit closure (the only sanctioned non-ingress keyring write); no long-lived Store.
- **Ingress**: `init --auth-code-stdin` two-phase OAuth; new `gro set-credential` (oauth_token, stdin/`--from-env`, allowed-key, value never echoed).
- **§1.6/§1.7**: `config show`/`--json`/`--verbose` + `clear`/`--all`/`--dry-run`; OAuth client JSON by path+presence+sha256-12; `config show`/`clear` use `OpenNoMigrate` (diagnostic/remediation), `config test` uses full `Open`.
- **§1.8 signal**: recorded in `internal/migrationsink` (leaf — no import cycle), spliced into JSON via the existing `output.JSON`, flushed to stderr by an execution-level `defer` in `root.runRoot` (above the `os.Exit` trap).
- **§1.12**: `internal/noleak` suite + hermetic `internal/credtest`.

## Out of scope (→ B2b)
Drive metadata cache relocation out of the config dir + auto-migration + `config clear --all` old-cache cleanup.

## Gates
[...]

PR #130, opened 2026-05-17
refactor: relocate Drive metadata cache to the OS cache dir (B2b)

### LINKED ISSUE #129: B2b: relocate Drive metadata cache to the OS cache dir (XDG)
Phase B unit **B2b** (orthogonal cache-hygiene slice; the security-complete migration was B2a / #127 / INT-438).

The Drive metadata cache currently lives **inside the config dir** (`$XDG_CONFIG_HOME/google-readonly/cache/drives.json`), conflating disposable cache with durable config. Relocate it to the OS-designated cache directory and migrate transparently.

### Scope
- Cache base dir → `os.UserCacheDir()/google-readonly` (Linux `$XDG_CACHE_HOME/...` or `~/.cache/...`; macOS `~/Library/Caches/...`; Windows `%LocalAppData%\...`).
- One-time transparent migration on cache open: carry over a warm `drives.json` if present, then remove the legacy `<configdir>/cache` dir (idempotent).
- `gro config clear --all` also clears the Drive cache (new location); legacy dir cleaned via the migration shim for one deprecation cycle.
- Hermetic cache tests (control `XDG_CACHE_HOME`/`HOME`); migration + idempotency tests; docs/help sweep.

### Out of scope
No secret handling, no auth changes (all delivered in B2a). Cache contents are disposable Drive listings.

Parent epic: INT-310. Jira: INT-XXX (linked after creation).

rianjs:
Phase B unit **B2b** (orthogonal cache-hygiene slice; security-complete migration was B2a / #128 / `e8699d2`). Closes #129. [INT-439]

## What
Relocate the Drive metadata cache out of the config dir into the OS-designated cache dir, with transparent one-time migration. No secret/auth changes.

- **Location:** `os.UserCacheDir()/google-readonly` — Linux `$XDG_CACHE_HOME` or `~/.cache`, macOS `~/Library/Caches`, Windows `%LocalAppData%`. Stdlib only, zero new deps. (Deliberate: canonical stdlib root, no Windows `\Cache` suffix — architect-reviewed/accepted.)
- **config:** pure `configDirPath()`/`CacheDirPath()` resolvers split from the creating `GetConfigDir()`/`GetCacheDir()` wrappers; `LegacyCacheDir()` + local `legacyCacheSubdir` (no `config`->`cache` import).
- **cache.New():** transparent, silent, best-effort one-time relocation (carry warm `drives.json`, then remove legacy dir). Never fails `New()` (cache is disposable); skips removal if the carry failed; idempotent.
- **config clear --all:** also clears the Drive cache and force-cleans the legacy dir; `--dry-run` uses the non-creating resolver and touches nothing; plain `clear` stays token-only.
- **tests:** hermetic cache/config tests (also fixes a pre-existing real-dir pollution bug — `cache_test.go` had no env isolation); migration/idempotency/clear coverage; de-brittled `TestCache_GetDir`; `credtest.Setup` isolates the cache dir too.
- **docs:** README cache location + auto-migration + `clear --all`.

## Review
Architect review converged 0/0/0/0 over 3 iterations before implementation; all findings folded into the plan (#129).

## Gates
gofmt clean, build, vet, `go test -race ./...` 1387 pass / 32 pkgs, golangci-lint v2.12.2 = 0, CI-pin v2.0.2 = 0.

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -13,8 +13,6 @@ import (
 const (
 	// DefaultTTLHours is the default cache TTL if not configured
 	DefaultTTLHours = 24
-	// CacheDir is the subdirectory within config for cache files
-	CacheDir = "cache"
 	// DrivesFile is the cache file for shared drives
 	DrivesFile = "drives.json"
 )
@@ -38,17 +36,17 @@ type Cache struct {
 	ttlHours int
 }
 
-// New creates a new Cache instance
+// New creates a new Cache instance rooted at the OS cache dir (B2b). It also
+// runs a transparent, best-effort one-time relocation of a pre-B2b cache that
+// lived inside the config dir; relocation never fails New (the cache is
+// disposable — it simply repopulates).
 func New(ttlHours int) (*Cache, error) {
-	configDir, err := config.GetConfigDir()
+	cacheDir, err := config.GetCacheDir()
 	if err != nil {
 		return nil, err
 	}
 
-	cacheDir := filepath.Join(configDir, CacheDir)
-	if err := os.MkdirAll(cacheDir, config.DirPerm); err != nil {
-		return nil, err
-	}
+	migrateLegacyCacheDir(cacheDir)
 
 	if ttlHours <= 0 {
 		ttlHours = DefaultTTLHours
@@ -58,6 +56,33 @@ func New(ttlHours int) (*Cache, error) {
 		dir:      cacheDir,
 		ttlHours: ttlHours,
 	}, nil
+}
+
+// migrateLegacyCacheDir relocates a pre-B2b cache (a "cache" subdir of the
+// config dir) into newDir, then removes the legacy dir. Strictly silent and
+// best-effort: any failure is abandoned without touching New's result. If the
+// warm-cache carry fails the legacy dir is left intact (don't delete data we
+// could not carry); a stuck legacy dir is force-cleaned by `config clear
+// --all`. Idempotent: once the legacy dir is gone this is a single stat.
+func migrateLegacyCacheDir(newDir string) {
+	legacy, err := config.LegacyCacheDir()
+	if err != nil {
+		return
+	}
+	if _, err := os.Stat(legacy); err != nil {
+		return // absent (steady state) or unreadable — nothing safe to do
+	}
+
+	legacyDrives := filepath.Join(legacy, DrivesFile)
+	newDrives := filepath.Join(newDir, DrivesFile)
+	if _, err := os.Stat(newDrives); os.IsNotExist(err) {
+		if data, rerr := os.ReadFile(legacyDrives); rerr == nil { //nolint:gosec // G304: path from config dir, not user input
+			if werr := os.WriteFile(newDrives, data, config.TokenPerm); werr != nil { //nolint:gosec // G703: config-derived cache path, user's own disposable data
+				return // carry failed: keep legacy intact, try again next run
+			}
+		}
+	}
+	_ = os.RemoveAll(legacy) // best-effort; cosmetic if it lingers
 }
 
 // GetDrives returns cached shared drives, or nil if cache is stale or missing
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -75,11 +75,18 @@ func migrateLegacyCacheDir(newDir string) {
 
 	legacyDrives := filepath.Join(legacy, DrivesFile)
 	newDrives := filepath.Join(newDir, DrivesFile)
-	if _, err := os.Stat(newDrives); os.IsNotExist(err) {
-		if data, rerr := os.ReadFile(legacyDrives); rerr == nil { //nolint:gosec // G304: path from config dir, not user input
+	if _, serr := os.Stat(newDrives); os.IsNotExist(serr) {
+		// New cache absent: try to carry the warm legacy file.
+		data, rerr := os.ReadFile(legacyDrives) //nolint:gosec // G304: path from config dir, not user input
+		switch {
+		case rerr == nil:
 			if werr := os.WriteFile(newDrives, data, config.TokenPerm); werr != nil { //nolint:gosec // G703: config-derived cache path, user's own disposable data
-				return // carry failed: keep legacy intact, try again next run
+				return // carry failed: keep legacy intact, retry next run
 			}
+		case os.IsNotExist(rerr):
+			// No warm file to carry — fine, fall through to cleanup.
+		default:
+			return // legacy drives file exists but unreadable: do NOT delete it
 		}
 	}
 	_ = os.RemoveAll(legacy) // best-effort; cosmetic if it lingers
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -75,7 +75,10 @@ func migrateLegacyCacheDir(newDir string) {
 
 	legacyDrives := filepath.Join(legacy, DrivesFile)
 	newDrives := filepath.Join(newDir, DrivesFile)
-	if _, serr := os.Stat(newDrives); os.IsNotExist(serr) {
+	switch _, serr := os.Stat(newDrives); {
+	case serr == nil:
+		// New cache already present: nothing to carry, safe to drop legacy.
+	case os.IsNotExist(serr):
 		// New cache absent: try to carry the warm legacy file.
 		data, rerr := os.ReadFile(legacyDrives) //nolint:gosec // G304: path from config dir, not user input
 		switch {
@@ -88,6 +91,8 @@ func migrateLegacyCacheDir(newDir string) {
 		default:
 			return // legacy drives file exists but unreadable: do NOT delete it
 		}
+	default:
+		return // ambiguous stat on the new path: do not risk deleting un-carried legacy
 	}
 	_ = os.RemoveAll(legacy) // best-effort; cosmetic if it lingers
 }
--- a/internal/cmd/config/config.go
+++ b/internal/cmd/config/config.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/open-cli-collective/cli-common/credstore"
 
+	"github.com/open-cli-collective/google-readonly/internal/cache"
 	"github.com/open-cli-collective/google-readonly/internal/config"
 	"github.com/open-cli-collective/google-readonly/internal/gmail"
 	"github.com/open-cli-collective/google-readonly/internal/keychain"
@@ -69,18 +70,15 @@ func newClearCommand() *cobra.Command {
 		Use:   "clear",
 		Short: "Remove the stored OAuth token (active profile)",
 		Long: `Remove the stored OAuth token under the active credential_ref,
-forcing re-authentication (§1.7). --all also removes config.yml. --dry-run
-reports what would be removed without removing it. The OAuth client JSON
-(deployment material) is never removed.
-
-Note: the Drive metadata cache is not yet relocated/cleared here — that
-lands in a follow-up unit (B2b).`,
+forcing re-authentication (§1.7). --all also removes config.yml and the Drive
+metadata cache. --dry-run reports what would be removed without removing it.
+The OAuth client JSON (deployment material) is never removed.`,
 		Args: cobra.NoArgs,
 		RunE: func(_ *cobra.Command, _ []string) error {
 			return runClear(all, dryRun)
 		},
 	}
-	cmd.Flags().BoolVar(&all, "all", false, "Also remove config.yml (active-profile scope)")
+	cmd.Flags().BoolVar(&all, "all", false, "Also remove config.yml and the Drive metadata cache")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "Report what would be removed; remove nothing")
 	return cmd
 }
@@ -239,6 +237,10 @@ func runClear(all, dryRun bool) error {
 		}
 		if all {
 			fmt.Printf("Would remove: %s\n", config.ShortenPath(cfgPath))
+			// Non-creating resolver: --dry-run must not create or migrate.
+			if cacheDir, cerr := config.CacheDirPath(); cerr == nil {
+				fmt.Printf("Would remove: Drive metadata cache at %s\n", config.ShortenPath(cacheDir))
+			}
 		}
 		fmt.Println()
 		fmt.Println("--dry-run: nothing was removed.")
@@ -262,6 +264,20 @@ func runClear(all, dryRun bool) error {
 			fmt.Printf("No %s to remove.\n", config.ShortenPath(cfgPath))
 		default:
 			return fmt.Errorf("removing %s: %w", config.ShortenPath(cfgPath), err)
+		}
+
+		// Drive metadata cache: cache.New also runs the transparent legacy
+		// relocation, then Clear removes the (new) cache dir. An explicit
+		// full reset additionally force-cleans the legacy dir even if the
+		// transparent shim had skipped it (carry-failed). All best-effort —
+		// the cache is disposable.
+		if c, cerr := cache.New(config.GetCacheTTLHours()); cerr == nil {
+			if clrErr := c.Clear(); clrErr == nil {
+				fmt.Printf("Removed Drive metadata cache at %s.\n", config.ShortenPath(c.GetDir()))
+			}
+		}
+		if legacy, lerr := config.LegacyCacheDir(); lerr == nil {
+			_ = os.RemoveAll(legacy)
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


rianjs:
**Major**
- `internal/cache/cache.go:79`: failed legacy carry can still delete the legacy cache. The comment and converged plan say “if the warm-cache carry fails the legacy dir is left intact,” but the implementation only returns on `WriteFile` failure. If `os.ReadFile(legacyDrives)` fails for anything other than absence, the code falls through to `os.RemoveAll(legacy)` at `internal/cache/cache.go:85`. That violates the agreed failure semantics and can delete data that was not carried. Treat `os.IsNotExist(rerr)` as “no warm file, okay to remove legacy dir,” but return on other read errors. Add a focused test by making `legacy/drives.json` a directory or otherwise unreadable and asserting `New()` still succeeds while legacy remains.

**Minor**
- `internal/cmd/config/config.go:227`: `--dry-run --all` still calls `config.GetConfigPath()` before the dry-run branch, and `GetConfigPath()` routes through `GetConfigDir()` via `inDir`, which creates the config directory. This is pre-existing and not cache migration, but it conflicts with the PR/test wording that dry-run “touches nothing” / “creates and removes nothing.” Either use a pure config path resolver for this display path or narrow the wording/tests to “does not create or migrate cache state.”

Everything else matches the converged B2b architecture: OS cache root is stdlib `UserCacheDir()/google-readonly`, pure/creating cache resolvers are split, migration is silent/non-fatal, `config clear --all` force-cleans cache while plain `clear` stays token-only, tests isolate `LOCALAPPDATA`, and `internal/config` does not import `internal/cache`.

rianjs:
## TDD Coverage Assessment

Independent review of the diff (branch `refactor/INT-439-drive-cache-xdg`). Local `go test -race ./internal/cache/... ./internal/config/... ./internal/cmd/config/...` → **all pass, 0 skips** (104 tests / 3 pkgs).

### Verdict: **Adequate** — coverage is solid and genuinely behavior-driven. A few weak spots noted, none blocking.

Every new/changed behavior has a regression-protecting test:

- **Resolver split** (`config_test.go` `TestCacheDirResolvers`): `GetCacheDir` creates + asserted under `os.UserCacheDir()/DirName` + asserted *not* under config tree; `CacheDirPath`/`LegacyCacheDir` asserted non-creating via `os.Stat`+`IsNotExist`; `GetConfigDir` still creates. `os.UserCacheDir()` computed in-test, no hard-coded per-OS strings. Good.
- **Migration semantics** (`cache_test.go` `TestMigrateLegacyCacheDir`): warm-carry + legacy removal + idempotency (second `New`); "new present ⇒ not overwritten, legacy still removed"; **carry-failed ⇒ legacy preserved + `New` still ok** (drives.json-as-dir forces a non-IsNotExist `ReadFile` error — exercises the real `return` branch at cache.go:84); "no legacy" no-op. The full matrix is covered.
- **clear semantics** (`behavior_test.go` `TestRunClearSemantics`): `--all` removes new + legacy; plain `clear` neither inits nor migrates (asserts new dir absent *and* legacy untouched); `--dry-run --all` creates/removes nothing and names the cache.
- **Hermeticity**: `hermetic`/`hermeticCfg` set HOME, XDG_CACHE_HOME, LOCALAPPDATA, XDG_CONFIG_HOME under `t.TempDir()`; `credtest.Setup` also isolates the cache dir. The pre-existing real-dir pollution bug in `cache_test.go` is fixed. `TestCacheDirResolvers` correctly gives each subtest its *own* env (creating resolver would otherwise defeat a sibling's non-creating assertion). No reliance on the developer's real dirs.

### Weak spots

1. **Suggestion — `GetConfigPathNoCreate` in `runClear` is not directly asserted as non-creating.** PR objective lists `GetConfigPathNoCreate` among the pure resolvers. `TestCacheDirResolvers` covers `CacheDirPath`/`LegacyCacheDir`/`GetConfigDir` but not `GetConfigPathNoCreate`. The "plain clear" subtest only indirectly exercises it (and only because no token path forces dir creation elsewhere). A direct `CacheDirPath`-style stat assertion on `GetConfigPathNoCreate` would lock this resolver's contract.

2. **Suggestion — `--all` clear never exercises the carry-failed force-clean *interaction*.** `behavior_test.go:--all` seeds an *empty* legacy dir (no `drives.json`), so `cache.New`'s shim carries nothing and removes legacy itself; the explicit `os.RemoveAll(legacy)` fallback in `runClear` is then redundant in that test. The scenario the fallback exists for — shim skipped removal because the carry failed, `--all` must still force-clean — is not reproduced (would need a legacy `drives.json` that fails to carry, e.g. drives.json-as-dir, then assert `--all` still nukes legacy). Currently a regression that broke the `runClear` fallback while leaving the shim intact would not be caught by this suite.

3. **Nit — `TestMigrateLegacyCacheDir` subtests share one parent-scoped `legacy`/`newDir` and rely on ordered cleanup.** Subtests run sequentially and each `defer c.Clear()` tidies the new dir, and the "unreadable" subtest leaves `legacy` intact by design but is last-but-one; the trailing "no legacy" subtest would observe that leftover `legacy` dir. It still passes (a stale legacy dir is a valid no-op precondition there), but the ordering coupling is implicit/fragile rather than each subtest fully isolating its own fixtures. Low risk; flagging for hygiene.

4. **Nit — carry-failed test asserts only legacy survival, not new-dir absence.** `unreadable legacy drives file` asserts legacy is preserved + `New` ok, but does not assert that `drives.json` was *not* written into the new dir (i.e. no partial/corrupt carry). The implementation is correct (write is gated on a successful `ReadFile`), but the test doesn't pin "no partial carry."

### Strengths worth noting

- De-brittled `TestCache_GetDir`: substring `Contains(..., "cache")` replaced with exact `Equal(c.GetDir(), config.CacheDirPath())` — correctly addresses the macOS `~/Library/Caches` / Windows no-`\Cache`-suffix portability gap. Right call.
- "Migration never fails `New`" is asserted in three independent subtests via `testutil.NoError(t, err)` after `New`. The non-fatal contract is well pinned.
- Assertions are mostly exact (`Equal`, `IsNotExist`, `Len`) rather than substring; `--dry-run` output check is a reasonable `Contains` on a stable label.

No flaky/shared-mutable-global risk beyond Nit 3; `t.Setenv` + `t.TempDir` give per-test isolation and `-race` is clean.

rianjs:
**Codex PR review + TDD assessment addressed** (commit `ebb90f6`):

Codex (1 major, 1 minor):
- **Major** — `migrateLegacyCacheDir`: a non-`IsNotExist` `ReadFile` error on the legacy drives file used to fall through to `RemoveAll`. Now the carry-failed path strictly preserves the legacy dir (remove only when the new cache is present, there is genuinely no warm file, or the carry succeeded). Focused test added (legacy `drives.json` as a directory) asserting legacy survives + no partial carry.
- **Minor** — added `config.GetConfigPathNoCreate`; `runClear` uses it so `clear` (incl. `--dry-run`) no longer creates the config dir as a side effect; `os.Remove` already tolerates an absent dir/file.

TDD (2 suggestions, 2 nits — all taken):
- direct non-creating assertion for `GetConfigPathNoCreate`;
- `--all` cache test now seeds a carry-failing legacy so `runClear`'s explicit `RemoveAll(legacy)` fallback is exercised (not just the shim);
- `TestMigrateLegacyCacheDir` subtests fully isolate their own hermetic env/fixtures (no ordered-cleanup coupling);
- carry-failed test also asserts no partial/corrupt carry into the new dir.

Gates: gofmt/build/vet, `go test -race ./...` 1388 pass / 32 pkgs, golangci-lint v2.12.2 = 0, CI-pin v2.0.2 = 0. The consolidated final Codex pass will run after the daemon review (quota-conserving batching).

internal/cache/cache.go
```diff
@@ -60,6 +58,40 @@ func New(ttlHours int) (*Cache, error) {
...
+		return
+	}
+	if _, err := os.Stat(legacy); err != nil {
+		return // absent (steady state) or unreadable — nothing safe to do
+	}
+
+	legacyDrives := filepath.Join(legacy, DrivesFile)
+	newDrives := filepath.Join(newDir, DrivesFile)
+	if _, serr := os.Stat(newDrives); os.IsNotExist(serr) {
+		// New cache absent: try to carry the warm legacy file.
```
monit-reviewer, line 82: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): If os.Stat(newDrives) returns a non-nil, non-IsNotExist error (e.g. permission denied), the carry block is skipped but execution falls through to os.RemoveAll(legacy), silently deleting the legacy warm cache without having carried it. The stated invariant is 'skip removal if the carry failed' — an explicit early-return on the ambiguous-stat case would enforce it. Note: the general ReadFile failure path was addressed in ebb90f6, but the Stat error path remains.

internal/cmd/config/config.go
```diff
@@ -263,6 +267,20 @@ func runClear(all, dryRun bool) error {
 		default:
 			return fmt.Errorf("removing %s: %w", config.ShortenPath(cfgPath), err)
 		}
+
```
monit-reviewer, line 269: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): runClear --all calls cache.New() which triggers migrateLegacyCacheDir before clearing. If c.Clear() fails silently, legacy data will have been migrated and left in the new cache dir while the legacy dir is force-removed — net result: data the user asked to clear ends up migrated-but-not-cleared with no diagnostic output. Using config.CacheDirPath() + os.RemoveAll directly in the --all path would avoid the migration side-effect and the spurious os.MkdirAll from GetCacheDir().

internal/cmd/config/config.go
```diff
@@ -263,6 +267,20 @@ func runClear(all, dryRun bool) error {
 		default:
```
monit-reviewer, line 266: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): Silent failure when clearing the Drive cache on --all: if cache.New() or c.Clear() fails, the user receives no output and no error — they may believe the full reset succeeded when the cache dir still exists. A best-effort warning to stderr on failure would match user expectations for an explicit destructive reset.

internal/cache/cache.go
```diff
@@ -60,6 +58,40 @@ func New(ttlHours int) (*Cache, error) {
...
+		data, rerr := os.ReadFile(legacyDrives) //nolint:gosec // G304: path from config dir, not user input
+		switch {
+		case rerr == nil:
+			if werr := os.WriteFile(newDrives, data, config.TokenPerm); werr != nil { //nolint:gosec // G703: config-derived cache path, user's own disposable data
+				return // carry failed: keep legacy intact, retry next run
```
monit-reviewer, line 87: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): The nolint comment cites G703, which is not a valid gosec rule ID. The suppressed check for os.WriteFile with a variable path would be G306 (file permissions) or G304 (tainted file path). Using a non-existent rule ID makes the rationale unverifiable and may silently suppress unrelated future warnings.

monit-reviewer (approved):
## Automated PR Review



**Reviewed commit:** `ebb90f6`

### Summary

| Reviewer | Findings |
|----------|----------|

<details>

**💡 Suggestion** - `internal/cmd/config/config.go:270`
> runClear --all calls cache.New() which triggers migrateLegacyCacheDir before clearing. If c.Clear() fails silently, legacy data will have been migrated and left in the new cache dir while the legacy dir is force-removed — net result: data the user asked to clear ends up migrated-but-not-cleared with no diagnostic output. Using config.CacheDirPath() + os.RemoveAll directly in the --all path would avoid the migration side-effect and the spurious os.MkdirAll from GetCacheDir().

</details>

<details>

**💡 Suggestion** - `internal/cache/cache.go:79`
> If os.Stat(newDrives) returns a non-nil, non-IsNotExist error (e.g. permission denied), the carry block is skipped but execution falls through to os.RemoveAll(legacy), silently deleting the legacy warm cache without having carried it. The stated invariant is 'skip removal if the carry failed' — an explicit early-return on the ambiguous-stat case would enforce it. Note: the general ReadFile failure path was addressed in ebb90f6, but the Stat error path remains.

</details>

<details>

**💡 Suggestion** - `internal/cmd/config/config.go:267`
> Silent failure when clearing the Drive cache on --all: if cache.New() or c.Clear() fails, the user receives no output and no error — they may believe the full reset succeeded when the cache dir still exists. A best-effort warning to stderr on failure would match user expectations for an explicit destructive reset.

</details>

<details>

**💡 Suggestion** - `internal/cache/cache.go:84`
> The nolint comment cites G703, which is not a valid gosec rule ID. The suppressed check for os.WriteFile with a variable path would be G306 (file permissions) or G304 (tainted file path). Using a non-existent rule ID makes the rationale unverifiable and may silently suppress unrelated future warnings.

</details>

*5 info-level observations excluded. Run with `--verbose` to include.*

*3 PR discussion threads considered.*

---
<details>
<summary>Completed in 2m 14s | $0.84 | sonnet | daemon 0.2.120 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Reviewers | hybrid-synthesis, documentation:docs-reviewer, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 2m 14s wall · 7m 22s compute (Reviewers: 1m 27s · Synthesis: 44s) |
| Cost | $0.84 |
| Tokens | 134.9k in / 30.5k out |
| Turns | 7 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 30.4k | 2.1k | 0 | 30.4k (1h) | $0.15 |
| documentation:docs-reviewer | haiku | 5.1k | 4.9k | 0 | 5.1k (1h) | $0.03 |
| harness-engineering:harness-architecture-reviewer | sonnet | 20.0k | 4.3k | 2.1k | 17.8k (1h) | $0.14 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 20.0k | 5.6k | 2.1k | 17.8k (1h) | $0.16 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 20.0k | 4.2k | 2.1k | 17.8k (1h) | $0.14 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 19.6k | 4.5k | 2.1k | 17.5k (1h) | $0.14 |
| security:security-code-auditor | haiku | 19.9k | 5.1k | 0 | 19.9k (1h) | $0.06 |

</details>

internal/cache/cache.go
rianjs: Fixed in 9535732. \`migrateLegacyCacheDir\` now switches on the \`os.Stat(newDrives)\` result: \`nil\` → new present, safe to drop legacy; \`IsNotExist\` → attempt the carry; **any other (ambiguous) stat error → early return**, so the "skip removal unless the carry is safe" invariant holds for the stat-error path too, not just the ReadFile path.

internal/cmd/config/config.go
rianjs: Fixed in 9535732. \`runClear --all\` no longer calls \`cache.New()\`; it \`os.RemoveAll\`s the current cache dir (\`config.CacheDirPath()\`) and the legacy dir directly — no migrate-then-delete, no \`MkdirAll\` side-effect, and the migration shim is never invoked on a clear. Removal errors are surfaced (see next).

internal/cmd/config/config.go
rianjs: Fixed in 9535732. With the direct \`os.RemoveAll\` in \`--all\`, a non-nil removal error is now returned (\`removing Drive metadata cache %s: %w\` / \`removing legacy Drive cache %s: %w\`) instead of silently swallowed — an explicit full reset can no longer falsely report success while the cache dir survives.

internal/cache/cache.go
rianjs: Respectfully declined — \`G703\` is a real gosec rule in the version that fires here. golangci-lint v2.12.2 emits literally \`G703: Path traversal via taint analysis (gosec)\` on this line (gosec added the G70x taint-analysis rules in recent versions; the CI-pinned v2.0.2 predates them, which is exactly why these are pre-emptive annotations — same pattern as the B2a keychain/migrate code). \`//nolint:gosec\` scoping is linter-level by golangci-lint design regardless of the trailing text; the \`G703:\` prefix is accurate documentation of the specific rule. Changing it to G304/G306 would name a rule that does not fire here.

rianjs:
No findings.

The current diff matches the converged INT-439/B2b architecture and product intent: cache moves to `os.UserCacheDir()/google-readonly`, resolver creation semantics are split cleanly, `cache.New()` migration is silent/non-fatal and preserves legacy on uncertain carry, `config clear --all` now clears current and legacy cache dirs directly without invoking migration, plain `config clear` remains token-only, dry-run uses non-creating path resolvers, tests are hermetic across Linux/macOS/Windows env knobs, and `internal/config` does not import `internal/cache`.

The retained `G703` nolint rationale is acceptable given the stated newer local gosec behavior and CI-pin context.

PR #134, opened 2026-05-19
feat(state): port gro to cli-common state components [MON-5371]

### LINKED ISSUE #131: Port gro to cli-common state components (statedir config + cli-common/cache; drop user-TTL & config cache cmds)
**working-with-state.md §6.4 unit 3 — gro (independent single-scope unit).** gro's cache *location* is already conformant (B2b `os.UserCacheDir()/google-readonly`) — this is adoption + hardening + command-surface removal, not a cache relocation.

## Scope (per-surface commits, §6.2)
- **Config:** `internal/config` hand-rolled `~/.config/google-readonly` (all OSes) → `cli-common/statedir` `Scope{Name:"google-readonly"}`. macOS/Windows relocation (Linux unchanged). Config write `os.WriteFile`→atomic; perms dir `0700`/file `0600`. Keep the no-create split (`config clear --dry-run` etc.) — map onto statedir `ConfigDir` vs `ConfigDirEnsured`.
- **Cache:** adopt `cli-common/cache` (Envelope[T], atomic temp-file-rename, version+identity-mismatch-as-miss, Classify/Age) with `Locator{Root: statedir.Cache{Tool:"google-readonly"}.CacheDir(), InstanceKey:"default"}`. gro currently uses plain `os.WriteFile` for cache — atomic is a real robustness win.
- **TTL:** remove user-configurable `cache_ttl_hours`; TTL becomes hard-coded per resource (§4.4). **Remove `gro config cache ttl|show|clear`** (clearing folds into `config clear --all`; `refresh`/`refresh --status` are sufficient).
- Replace gro's per-CLI `credtest`/XDG isolation with `cli-common/statedirtest` for state tests (it's missing USERPROFILE/AppData/XDG_DATA_HOME today — a real Windows leak).

## §3.2 acceptance matrix (merge gate)
Declare copy-vs-move / second-run / downgrade-fork. Prove all 8 cases against gro's real legacy sources (legacy `config.json`, the B2b cache). Conflicting old vs new config ⇒ fail loud.

## Out of scope
- Secrets/keyring (already cli-common/credstore via B2a — location-independent).
- Cutting the cli-common semver tag (closing ticket).

Depends on commons (MON-5364 / cli-common#18, merged). Order per §6.4.

Jira: MON-5371 (child of INT-310, KTLO)

rianjs:
Closes #131. Implements the gro half of working-with-state.md §6.4 unit 3:
config-dir resolver via cli-common/statedir; atomic config writes; one-time
copy-leave-old relocation on macOS/Windows; cli-common/cache envelope with
hard-coded per-resource TTL; removal of the user-configurable cache TTL knob
and the \`gro config cache\` subcommand.

## §3.2 acceptance matrix (per merge gate)

**Config policy:** copy-leave-old at \`gro init\`; runtime mutation-free w/
ErrRelocationConflict soft-fallback (\`LoadConfigForRuntime\`); divergent
old/new fails loud naming both paths and mutating nothing. Linux: zero-byte
op (old path ≡ new path). \`token.json\` is excluded from the copy and
handled exclusively by the existing §1.8 token migrator, which now probes
both old and new paths with full conflict semantics.

**Cache policy:** drop-on-floor. The new cli-common envelope schema makes
the old \`DriveCache\` shape an identity/version-mismatch miss; a malformed
envelope is mapped to a miss inside gro's wrapper (cache is disposable per §4).

| Case | Config | Cache |
|---|---|---|
| old-only | TestRelocate_OldOnly_CopiedAtInit | TestMigrateLegacyCacheDir/carries warm cache |
| new-only | TestRelocate_NewOnly_LeftUntouched | TestCache_GetSetDrives |
| same value both | TestRelocate_Equal_NoOp | TestCache_CorruptedCache/old DriveCache shape → miss |
| **conflicting** | TestRelocate_Divergent_FailLoudNamesBothPaths_MutatesNothing | (disposable: schema-mismatch → miss; covered above) |
| malformed old | TestRelocate_MalformedOld_FailLoud | TestMigrateLegacyCacheDir/unreadable legacy preserved |
| malformed new | TestRelocate_MalformedNew_FailLoud | TestCache_CorruptedCache/malformed JSON treated as miss |
| neither | TestRelocate_Neither_PathResolvedNotCreated | TestCache_GetSetDrives/returns nil for missing |
| no real-dir writes | statedirtest.Hermetic everywhere | statedirtest.Hermetic |

## Surfaces

1. **Config dir resolver**: \`configDirPath()\` → \`statedir.Scope{Name:\"google-readonly\"}.ConfigDir()/ConfigDirEnsured()\`.
2. **Atomic SaveConfig**: \`os.CreateTemp\` + \`Chmod 0600\` + \`Rename\`.
3. **Relocation gate**: new \`internal/config/relocate.go\` — \`DetectConfigRelocation\` (mutation-free), \`ApplyConfigRelocation\` (whole-dir copy minus token.json, atomic per-file, idempotent), \`LoadConfigForRuntime\` (soft-conflict wrapper for non-init callers). Init wires the gate ahead of \`EnsureMigrated\`.
4. **Cache adoption**: \`internal/cache/cache.go\` now wraps \`cli-common/cache.Locator{Root: statedir.Cache{Tool:\"google-readonly\"}.CacheDir(), InstanceKey:\"default\"}\` with \`ReadResource\`/\`WriteResource\` and hard-coded 24h TTL for drives.
5. **Drop user-TTL**: \`Config.CacheTTLHours\`, \`DefaultCacheTTLHours\`, \`GetCacheTTL{,Hours}\`, \`Prompter.AskCacheTTL\`, init step 6 TTL block — gone. Older config.yml with \`cache_ttl_hours:\` still loads (yaml ignores unknowns).
6. **Remove cache subcommand**: deleted \`gro config cache {show,clear,ttl}\` (clearing folds into \`gro config clear --all\`).
7. **Hermetic test helper**: \`credtest.Setup\` delegates to \`statedirtest.Hermetic\` (full 7-var set — closes a Windows real-dir leak). New \`credtest.ConfigDir(t)\` for tests that need the resolved dir.

## Token-secret hygiene

\`ApplyConfigRelocation\` deliberately excludes \`token.json\` — copying it would
leave a stale plaintext at the old path (the exact class MON-5370 guarded
against). Instead, \`keychain/migrate.go\` \`discover()\` was extended to
enumerate the old hand-rolled token path alongside the new one. Conflict
detection (\`planMigration\`) treats divergent old/new tokens as a §1.8
conflict and refuses to migrate without \`--overwrite\` — no side-channel scrub
can race ahead of the planner.

## Out of scope

- Secrets/keyring storage (location-independent via cli-common/credstore).
- Cutting the cli-common semver tag (blocked on MON-5375).
- Pre-B2b \`<configdir>/cache/\` migrator stays — its purpose is location, not
  schema, and it's still useful for installs that pre-date B2b.

## Verification

- \`make test\` ✅ 1368 tests pass.
- \`make lint\` ✅ 0 issues.
- Codex architect review (session \`019e4019\`, reused across the whole INT-310
  state-port batch) converged at blockers=0 majors=0 minors=0 nits=1 (cosmetic).


rianjs:
### Findings

**Blocker:** `oauth_client_path` is durable config but is excluded from relocation conflict detection.  
`internal/config/relocate.go:157-173` ignores `OAuthClientPath` entirely, even though `internal/config/config.go` documents it as an org/user override. That means old/new `config.yml` files with different explicit OAuth client paths are classified as equal and runtime silently uses the new side. The §3.2 conflicting-config row should fail loud for this. Treat old-default vs new-default as equivalent if needed, but explicit non-default divergence needs to conflict.

**Blocker:** old-only `config.json` installs are not relocated or read from the old hand-rolled dir.  
`detectRelocation` only checks `old/config.yml` (`internal/config/relocate.go:103-113`). `LoadConfig` only falls back to `old/config.yml` (`internal/config/config.go:247-274`), and `loadLegacyJSON` reads `config.json` only from the new statedir dir (`internal/config/config.go:285-300`). A pre-port macOS/Windows user with only old `config.json` is treated like a fresh install, so the existing per-file migrators never see the old config. This breaks the “whole-dir copy so existing migrators still work” plan.

**Major:** pre-B2b cache relocation does not cover the old hand-rolled config dir.  
`LegacyCacheDir` now resolves under the new statedir config dir (`internal/config/config.go:145-153`), while `ApplyConfigRelocation` skips directories (`internal/config/relocate.go:208-211`). So `<old hand-rolled>/google-readonly/cache/drives.json` is neither copied at init nor found by `migrateLegacyCacheDir` (`internal/cache/cache.go:75-108`). The PR matrix’s old-only cache carry test covers only the new-dir legacy-cache seam, not the macOS/Windows old-dir state.

**Minor:** the runtime soft-conflict test is a stub.  
`TestLoadConfigForRuntime_SoftConflict` sets up old/new divergent configs but never calls `LoadConfigForRuntime` or asserts returned canonical config / nil error / warning behavior (`internal/config/relocate_test.go:283-316`). Given runtime soft fallback is a key architectural distinction from init fail-loud, this should be pinned.

rianjs:
## TDD coverage assessment

**Verdict: Gaps**

The relocation gate, cache adoption, and init ordering invariant are well-covered. Two genuine gaps where a regression could ship green:

### Gap 1 — keychain `discover()` old-path candidate is untested
`internal/keychain/migrate.go:243-253` adds a new `OldHandRolledTokenPath()` candidate with a dedupe against the new path. There are no tests in `internal/keychain/` exercising `discover()` at all (no `migrate_test.go`), so:
- The macOS/Windows "old != new emits both candidates" behavior is unverified.
- The Linux "old == new must dedupe (no double candidate with identical bytes)" invariant is unverified.
- Worse: a regression that drops the new branch entirely (leaving only the new statedir path) would not be caught — and that would silently strand pre-MON-5371 token.json files for upgraders on macOS/Windows, exactly the platforms this PR's relocation is targeted at.

Suggest a `discover()` test with the legacy-scan env vars set, two `token.json` files seeded under injected old/new paths, asserting candidate count and locations across the path-divergence and path-identity cases.

### Gap 2 — `LoadConfigForRuntime` soft-conflict path doesn't actually assert behavior
`relocate_test.go:283-316` (`TestLoadConfigForRuntime_SoftConflict`) sets up divergent old/new, then never calls `LoadConfigForRuntime` — it `t.Skip`s on Linux and otherwise just falls off the end without invoking the SUT or checking the once-warn fires / canonical-new is returned. The soft-conflict semantic (warn once, return new cfg) is currently unproven; a regression that returned the old cfg, or panicked, or warned every call, would ship green.

Either inject a resolver seam so the case is exercised on Linux, or drop the test and add a unit test on `warnReloConflictOnce` plus a small refactor that makes the LoadConfigForRuntime branch table-testable without `t.Setenv` gymnastics.

Everything else (cache parse-error→miss, pre-5371 shape→miss, drop-on-floor Clear semantics, detect-runs-before-migrate ordering, copy-needed→apply ordering, idempotent apply, token.json excluded from copy) is adequately proven.

rianjs:
### Findings

**Blocker:** old-only malformed config still mutates before failing.  
`detectRelocation` now recognizes old `config.yml` or `config.json`, but in the `oldPresent && !newPresent` branch it returns `CopyNeeded` without parsing the old file first (`internal/config/relocate.go:121-128`). `gro init` then applies the relocation before `EnsureMigrated` (`internal/cmd/initcmd/init.go:290-299`), so an old-only malformed YAML/JSON file gets copied into the new dir before `LoadConfig` later fails parsing it (`internal/config/config.go:262-284`). That breaks the §3.2 malformed-old row: fail loud, mutate nothing. Please parse `oldPath` before returning `relocOldOnly`, and add old-only malformed YAML + JSON tests asserting the new dir/file is not created.

rianjs:
No findings. The old-only malformed branch now validates before `CopyNeeded`, preserving the fail-loud/mutate-nothing invariant for both YAML and JSON. The prior oauth path, config.json, cache relocation, runtime soft-conflict, and token-source concerns also look aligned with the batch architecture.

Focused verification run: `go test ./internal/config ./internal/cache ./internal/keychain` passed.

internal/config/relocate.go
```diff
@@ -0,0 +1,340 @@
...
+	if len(a) != len(b) {
+		return false
+	}
+	aa := append([]string(nil), a...)
+	bb := append([]string(nil), b...)
+	sort.Strings(aa)
+	sort.Strings(bb)
+	return reflect.DeepEqual(aa, bb)
+}
+
```
monit-reviewer, line 239: 🟡 **Medium** (harness-engineering:harness-enforcement-reviewer): configsMaterialEqual only checks Keyring.Backend. If KeyringConfig has additional fields (e.g. a passphrase env var for the file backend), two configs differing only on those fields would be classified as equal, silently ignoring a genuine user divergence. The equality check should use reflect.DeepEqual on the entire Keyring sub-struct, or explicitly enumerate every field.

internal/config/config.go
```diff
@@ -226,33 +219,78 @@ func ShortenPath(path string) string {
...
+		relErr = derr
+	} else if derr != nil && !errors.Is(derr, ErrRelocationConflict) {
+		return nil, derr
+	}
 
-	ymlPath, err := GetConfigPath()
-	if err != nil {
-		return nil, err
+	// Priority: new/config.yml → old/config.yml (only if new is absent) →
+	// legacy config.json at the new dir.
```
monit-reviewer: 🟡 **Medium** (harness-engineering:harness-enforcement-reviewer): When both old and new configs are present and the new config.yml is malformed, detectRelocation returns ErrRelocationConflict (wrapped), so LoadConfig sets relErr and continues. It then tries to re-parse the malformed YAML and returns a bare parse error. LoadConfigForRuntime does not recognize this as ErrRelocationConflict, so it propagates the hard parse error instead of soft-degrading. The test matrix covers detectRelocation in isolation but has no test exercising the full LoadConfig → LoadConfigForRuntime stack for the malformed-new, old-present scenario.

internal/cache/cache.go
```diff
@@ -1,196 +1,172 @@
...
-
 	return &Cache{
-		dir:      cacheDir,
-		ttlHours: ttlHours,
+		loc: clicache.Locator{Root: cacheDir, InstanceKey: instanceKey},
 	}, nil
 }
 
 // migrateLegacyCacheDir relocates a pre-B2b cache (a "cache" subdir of the
 // config dir) into newDir, then removes the legacy dir. Strictly silent and
```
monit-reviewer, line 68: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): migrateLegacyCacheDir now probes two legacy locations: LegacyCacheDir (statedir-resolved) and OldHandRolledLegacyCacheDir (pre-MON-5371 hand-rolled). The second probe is the one that matters on a macOS/Windows install that pre-dates both B2b and MON-5371, yet cache_test.go has no subtest that seeds a warm cache exclusively at the OldHandRolledLegacyCacheDir path and verifies it is carried. The dual-probe logic is new risky code with only one probe tested.

internal/config/relocate.go
```diff
@@ -0,0 +1,340 @@
...
+		return p, true
+	}
+	return "", false
+}
+
+// LoadConfigForRuntime is the soft-conflict variant of LoadConfig for non-init
+// callers. On ErrRelocationConflict it prints a one-shot stderr warning, then
+// returns the canonical (new-dir) config so the command can keep working.
+// Init uses strict LoadConfig (fail-loud) at its relocation gate.
+func LoadConfigForRuntime() (*Config, error) {
```
monit-reviewer, line 339: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): reloConflictOnce is a package-level sync.Once that is never reset between test runs in the same binary. Any test that exercises LoadConfigForRuntime in a conflict state will exhaust the Once, silently suppressing the warning for all subsequent tests in the same binary. If a future test ever asserts on the stderr warning, it will flake depending on execution order. Consider a reset hook or injecting an io.Writer so tests can verify the behavior independently.

internal/cache/cache.go
```diff
@@ -1,196 +1,172 @@
...
 
-// DriveCache represents the cached shared drives data
-type DriveCache struct {
-	CachedAt time.Time      `json:"cached_at"`
-	TTLHours int            `json:"ttl_hours"`
-	Drives   []*CachedDrive `json:"drives"`
-}
-
-// Cache provides TTL-based caching for API metadata
+// Cache is gro's wrapper around the cli-common envelope cache.
```
monit-reviewer, line 47: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): drivesTTL = "24h" is a string constant parsed via time.ParseDuration at every WriteResource call. A malformed string after a refactor surfaces only at the first cache write in production. Add a package-level init-time validation (e.g. panic on bad parse) or a TestMain assertion so the invariant is caught at test time.

internal/cache/cache.go
```diff
@@ -1,196 +1,172 @@
...
+	return os.RemoveAll(c.loc.Root)
 }
 
-// GetDir returns the cache directory path
+// GetDir returns the cache directory path.
 func (c *Cache) GetDir() string {
-	return c.dir
+	return c.loc.Root
 }
+
```
monit-reviewer, line 173: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): nowFn is an unexported package-level variable mutated in tests without a t.Parallel() guard. If anyone adds t.Parallel() to the outer test or its subtests, writes to nowFn from concurrent subtests will be a data race caught by -race. Prefer passing a clock func through Cache at construction time rather than a mutable global.

internal/cmd/initcmd/init.go
```diff
@@ -272,6 +280,25 @@ func readAllStdin() (string, error) {
 
 // runWith is the testable entry point for the wizard. NewCommand wraps it.
 func runWith(ctx context.Context, d initDeps, opts *initOptions) error {
+	// Step -1 (must precede the §1.8 migration): the MON-5371 config-dir
+	// relocation gate. If the old hand-rolled dir and the new statedir-
+	// resolved dir both exist with divergent settings, abort BEFORE
+	// EnsureMigrated runs — otherwise that path would soft-read one side via
+	// the runtime wrapper, migrate, then SaveConfig, papering over the
+	// conflict. On only-old we copy old→new (atomic per-file, skipping
+	// token.json which the §1.8 migrator handles).
```
monit-reviewer, line 289: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): The nil guard `if reloc.CopyNeeded && d.ApplyConfigRelocation != nil` silently skips the copy when ApplyConfigRelocation is nil but DetectConfigRelocation signalled CopyNeeded. A test stub injecting only DetectConfigRelocation can hit this path and leave the config unrelocated with no error or log. Either remove the nil guard and require ApplyConfigRelocation in initDeps, or return an error when CopyNeeded is true and the func is nil.

internal/cmd/initcmd/init.go
```diff
@@ -367,24 +394,13 @@ func runWith(ctx context.Context, d initDeps, opts *initOptions) error {
 	// Step 5: persist granted scopes (creates config.json if missing).
```
monit-reviewer, line 381: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): The configExistedBefore variable and its initialization are dead code suppressed with `_ = configExistedBefore` rather than removed. The snapshot-before-save gate it guarded (AskCacheTTL) is gone; the variable, its assignment in runWith, and the parameter in finishExisting with its blank-identifier suppressor should all be deleted to avoid misleading future readers.

monit-reviewer (dismissed):
## Automated PR Review



**Reviewed commit:** `7ba3105`

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:harness-architecture-reviewer | 3 |
| harness-engineering:harness-enforcement-reviewer | 3 |

<details>
<summary><strong>harness-engineering:harness-architecture-reviewer (3 findings)</strong></summary>

**💡 Suggestion** - `internal/config/relocate.go:325`
> reloConflictOnce is a package-level sync.Once that is never reset between test runs in the same binary. Any test that exercises LoadConfigForRuntime in a conflict state will exhaust the Once, silently suppressing the warning for all subsequent tests in the same binary. If a future test ever asserts on the stderr warning, it will flake depending on execution order. Consider a reset hook or injecting an io.Writer so tests can verify the behavior independently.

**💡 Suggestion** - `internal/cache/cache.go:47`
> drivesTTL = "24h" is a string constant parsed via time.ParseDuration at every WriteResource call. A malformed string after a refactor surfaces only at the first cache write in production. Add a package-level init-time validation (e.g. panic on bad parse) or a TestMain assertion so the invariant is caught at test time.

**💡 Suggestion** - `internal/cache/cache.go:170`
> nowFn is an unexported package-level variable mutated in tests without a t.Parallel() guard. If anyone adds t.Parallel() to the outer test or its subtests, writes to nowFn from concurrent subtests will be a data race caught by -race. Prefer passing a clock func through Cache at construction time rather than a mutable global.

</details>

<details>
<summary><strong>harness-engineering:harness-enforcement-reviewer (3 findings)</strong></summary>

**⚠️ Should Fix** - `internal/config/relocate.go:229`
> configsMaterialEqual only checks Keyring.Backend. If KeyringConfig has additional fields (e.g. a passphrase env var for the file backend), two configs differing only on those fields would be classified as equal, silently ignoring a genuine user divergence. The equality check should use reflect.DeepEqual on the entire Keyring sub-struct, or explicitly enumerate every field.

**⚠️ Should Fix** - `internal/config/config.go:248`
> When both old and new configs are present and the new config.yml is malformed, detectRelocation returns ErrRelocationConflict (wrapped), so LoadConfig sets relErr and continues. It then tries to re-parse the malformed YAML and returns a bare parse error. LoadConfigForRuntime does not recognize this as ErrRelocationConflict, so it propagates the hard parse error instead of soft-degrading. The test matrix covers detectRelocation in isolation but has no test exercising the full LoadConfig → LoadConfigForRuntime stack for the malformed-new, old-present scenario.

**💡 Suggestion** - `internal/cache/cache.go:68`
> migrateLegacyCacheDir now probes two legacy locations: LegacyCacheDir (statedir-resolved) and OldHandRolledLegacyCacheDir (pre-MON-5371 hand-rolled). The second probe is the one that matters on a macOS/Windows install that pre-dates both B2b and MON-5371, yet cache_test.go has no subtest that seeds a warm cache exclusively at the OldHandRolledLegacyCacheDir path and verifies it is carried. The dual-probe logic is new risky code with only one probe tested.

</details>

<details>

**💡 Suggestion** - `internal/cmd/initcmd/init.go:289`
> The nil guard `if reloc.CopyNeeded && d.ApplyConfigRelocation != nil` silently skips the copy when ApplyConfigRelocation is nil but DetectConfigRelocation signalled CopyNeeded. A test stub injecting only DetectConfigRelocation can hit this path and leave the config unrelocated with no error or log. Either remove the nil guard and require ApplyConfigRelocation in initDeps, or return an error when CopyNeeded is true and the func is nil.

</details>

<details>

**💡 Suggestion** - `internal/cmd/initcmd/init.go:394`
> The configExistedBefore variable and its initialization are dead code suppressed with `_ = configExistedBefore` rather than removed. The snapshot-before-save gate it guarded (AskCacheTTL) is gone; the variable, its assignment in runWith, and the parameter in finishExisting with its blank-identifier suppressor should all be deleted to avoid misleading future readers.

</details>

---
<details>
<summary>Completed in 4m 06s | $1.84 | sonnet | daemon 0.2.120 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Reviewers | hybrid-synthesis, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 4m 06s wall · 13m 04s compute (Reviewers: 3m 10s · Synthesis: 54s) |
| Cost | $1.84 |
| Tokens | 301.8k in / 45.1k out |
| Turns | 6 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 33.1k | 3.0k | 0 | 33.1k (1h) | $0.18 |
| harness-engineering:harness-architecture-reviewer | sonnet | 53.9k | 10.0k | 2.1k | 51.8k (1h) | $0.39 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 53.9k | 10.1k | 2.1k | 51.8k (1h) | $0.39 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 53.9k | 10.6k | 2.1k | 51.8k (1h) | $0.40 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 53.3k | 7.6k | 2.1k | 51.2k (1h) | $0.35 |
| security:security-code-auditor | haiku | 53.6k | 3.7k | 0 | 53.6k (1h) | $0.13 |

</details>

internal/config/relocate.go
rianjs: Fixed in the next push: `configsMaterialEqual` now uses `reflect.DeepEqual` on the whole `Keyring` sub-struct so any future field added to `KeyringConfig` is auto-covered.

internal/config/config.go
rianjs: Fixed: `LoadConfig` now skips the new-dir re-parse + legacy-JSON re-read when `relErr` is already set (a detect-time ErrRelocationConflict — which includes the malformed-new-wrapped-in-conflict case). The wrapped conflict error wins, so `LoadConfigForRuntime` soft-degrades correctly instead of hard-failing on a bare parse error.

internal/cache/cache.go
rianjs: Added `old-hand-rolled legacy cache subdir is carried (pre-MON-5371 install)` subtest to `TestMigrateLegacyCacheDir`: seeds the warm cache exclusively at `OldHandRolledLegacyCacheDir` and asserts it's carried and the legacy dir is removed.

internal/cmd/initcmd/init.go
rianjs: Fixed: `configExistedBefore` and its plumbing (variable + threading through `tryExistingToken` and `finishExisting`) are now deleted entirely, not suppressed with `_ =`.

internal/config/relocate.go
rianjs: Bugger off — the package-level `reloConflictOnce` is by design (one stderr warning per process lifetime is the user-facing contract). The runtime-soft-conflict test resets it via the testable seam (`loadConfigForRuntimeAt` sets `reloConflictOnce = sync.Once{}`). No test currently asserts on stderr capture; adding a hook for hypothetical future flake-prone tests is speculation we can address when (if) such a test is written.

internal/cache/cache.go
rianjs: Bugger off — `drivesTTL = "24h"` is a compile-time string constant in the same file as the only consumer; a typo there would surface in CI on the very next test run that touches the cache. The proposed init-time validation panics on a value that can't change between commits — it's mass-balanced overengineering for a non-issue.

internal/cache/cache.go
rianjs: Bugger off — `nowFn` is the standard test-clock seam pattern; no production caller uses `t.Parallel`, and the cache_test.go consumers don't either (they save/restore via defer). Adding a clock-injection constructor solely so a hypothetical future parallel test wouldn't race is overengineering: the moment a parallel test is actually written, the race detector flags it loudly and we refactor then.

internal/cmd/initcmd/init.go
rianjs: Bugger off — the nil guard is defense-in-depth for a degenerate test stub that registers Detect but not Apply (which would itself be a test bug, not a production path). Production `defaultDeps()` always sets both. Forcing an error on misconfigured stubs would be noisier without catching anything real; the test guard catches the realistic case (no relocation needed).

internal/config/relocate.go
monit-reviewer: **Settled**: monit-reviewer flagged that configsMaterialEqual only checked Keyring.Backend, potentially missing other KeyringConfig fields. rianjs fixed it to use reflect.DeepEqual on the whole Keyring sub-struct.

internal/config/config.go
monit-reviewer: **Settled**: monit-reviewer flagged that a malformed new config combined with an existing old config could cause LoadConfig to return a bare parse error instead of the expected ErrRelocationConflict. rianjs fixed LoadConfig to skip re-parsing when relErr is already set, so the wrapped conflict error propagates correctly.

internal/cache/cache.go
monit-reviewer: **Settled**: monit-reviewer flagged that the new OldHandRolledLegacyCacheDir probe in migrateLegacyCacheDir lacked test coverage. rianjs added a subtest seeding the warm cache exclusively at OldHandRolledLegacyCacheDir and asserting it is carried over.

internal/config/relocate.go
monit-reviewer: **Settled**: monit-reviewer flagged that the package-level reloConflictOnce could be exhausted across test runs. rianjs pushed back, explaining the once-per-process warning is intentional by design and that the testable seam already handles reset; no action taken.

internal/cache/cache.go
monit-reviewer: **Settled**: monit-reviewer suggested adding init-time validation for the drivesTTL string constant. rianjs declined, noting it's a compile-time constant in the same file as its only consumer and any typo would be caught immediately in CI.

internal/cache/cache.go
monit-reviewer: **Settled**: monit-reviewer suggested replacing the package-level nowFn test seam with constructor injection to avoid potential data races if parallel tests are added. rianjs declined, noting no parallel tests exist and the race detector would flag any future violation at that time.

internal/cmd/initcmd/init.go
monit-reviewer: **Settled**: monit-reviewer flagged that the nil guard on ApplyConfigRelocation could silently skip a needed copy with no error. rianjs declined, explaining the guard is defense-in-depth for degenerate test stubs and production always sets both functions via defaultDeps().

internal/cmd/initcmd/init.go
monit-reviewer: **Settled**: monit-reviewer flagged configExistedBefore as dead code suppressed with a blank identifier. rianjs fixed it by deleting the variable and all its plumbing entirely.

internal/config/config.go
```diff
@@ -226,33 +219,83 @@ func ShortenPath(path string) string {
 	return path
```
monit-reviewer, line 219: 🟡 **Medium** (harness-engineering:harness-architecture-reviewer): LoadConfig doc comment says the canonical new-dir config is returned alongside ErrRelocationConflict so callers can soft-degrade, but the implementation never reads the new-dir config when relErr != nil — the read block is guarded by `reloc.NewPath != "" && relErr == nil`. LoadConfigForRuntime therefore returns an all-defaults *Config (empty OAuthClientPath, empty GrantedScopes) instead of the actual new-dir settings. Consequences: (1) users with a non-default oauth_client_path get silent auth failures ('unable to read OAuth client JSON') during the conflict window; (2) the stale-scope check in me.go is silently skipped (CheckScopesMigration returns '' for an empty slice), masking a re-auth requirement. Fix: attempt the new-dir read unconditionally, store it separately from the relErr gate, and return it alongside the conflict error.

internal/config/relocate.go
```diff
@@ -0,0 +1,344 @@
...
+		return true
+	}
+	aIsDefault := aPath == "" || aPath == filepath.Join(aDir, OAuthClientFile)
+	bIsDefault := bPath == "" || bPath == filepath.Join(bDir, OAuthClientFile)
+	return aIsDefault && bIsDefault
+}
+
+func slicesEqualSorted(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
```
monit-reviewer, line 231: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): oauthClientPathEquiv compares OAuthClientPath values from loadConfigFromFile, which deliberately does not call applyDefaults (and therefore does not call ExpandPath). A user whose old config.yml has `oauth_client_path: ~/oauth_client.json` and whose new config.yml has the same value as a fully-expanded absolute path will see a false ErrRelocationConflict because the tilde form and the expanded form differ as strings. Both sides should be normalized (via ExpandPath or equivalent) before the equality comparison.

internal/cache/cache.go
```diff
@@ -1,196 +1,172 @@
...
+	tried := map[string]bool{}
+	for _, p := range probes {
+		legacy, err := p()
+		if err != nil || tried[legacy] {
+			continue
+		}
+		tried[legacy] = true
+		tryCarryLegacyCache(legacy, newDir)
 	}
+}
```
monit-reviewer, line 96: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): Cache.Clear() calls os.RemoveAll(c.loc.Root), which removes the entire tool-level cache root (os.UserCacheDir()/google-readonly), not just the current instance subtree. If a future cli-common Locator introduces a second instance key, any Clear() call silently evicts all instances. The safer scope is os.RemoveAll(filepath.Join(c.loc.Root, c.loc.InstanceKey)) to match only the instance being managed.

monit-reviewer (dismissed):
## Automated PR Review



**Reviewed commit:** `36c1355` | **Previous:** `7ba3105` (incremental)

### Summary

| Reviewer | Findings |
|----------|----------|

<details>

**⚠️ Should Fix** - `internal/config/config.go:219`
> LoadConfig doc comment says the canonical new-dir config is returned alongside ErrRelocationConflict so callers can soft-degrade, but the implementation never reads the new-dir config when relErr != nil — the read block is guarded by `reloc.NewPath != "" && relErr == nil`. LoadConfigForRuntime therefore returns an all-defaults *Config (empty OAuthClientPath, empty GrantedScopes) instead of the actual new-dir settings. Consequences: (1) users with a non-default oauth_client_path get silent auth failures ('unable to read OAuth client JSON') during the conflict window; (2) the stale-scope check in me.go is silently skipped (CheckScopesMigration returns '' for an empty slice), masking a re-auth requirement. Fix: attempt the new-dir read unconditionally, store it separately from the relErr gate, and return it alongside the conflict error.

</details>

<details>

**💡 Suggestion** - `internal/config/relocate.go:225`
> oauthClientPathEquiv compares OAuthClientPath values from loadConfigFromFile, which deliberately does not call applyDefaults (and therefore does not call ExpandPath). A user whose old config.yml has `oauth_client_path: ~/oauth_client.json` and whose new config.yml has the same value as a fully-expanded absolute path will see a false ErrRelocationConflict because the tilde form and the expanded form differ as strings. Both sides should be normalized (via ExpandPath or equivalent) before the equality comparison.

</details>

<details>

**💡 Suggestion** - `internal/cache/cache.go:96`
> Cache.Clear() calls os.RemoveAll(c.loc.Root), which removes the entire tool-level cache root (os.UserCacheDir()/google-readonly), not just the current instance subtree. If a future cli-common Locator introduces a second instance key, any Clear() call silently evicts all instances. The safer scope is os.RemoveAll(filepath.Join(c.loc.Root, c.loc.InstanceKey)) to match only the instance being managed.

</details>

*12 PR discussion threads considered.*

---
<details>
<summary>Completed in 4m 03s | $1.72 | sonnet | daemon 0.2.120 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Mode | Re-review · Cycle 2 · Session resumed |
| Reviewers | hybrid-synthesis, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 4m 03s wall · 10m 31s compute (Reviewers: 2m 49s · Synthesis: 47s) |
| Cost | $1.72 |
| Tokens | 334.0k in / 34.5k out |
| Turns | 6 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 42.2k | 2.6k | 13.6k | 28.5k (1h) | $0.15 |
| harness-engineering:harness-architecture-reviewer | sonnet | 54.0k | 8.5k | 2.1k | 51.9k (1h) | $0.37 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 54.0k | 7.2k | 2.1k | 51.9k (1h) | $0.35 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 54.0k | 6.2k | 2.1k | 51.9k (1h) | $0.33 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 53.3k | 7.2k | 2.1k | 51.2k (1h) | $0.35 |
| security:security-code-auditor | haiku | 53.6k | 1.9k | 0 | 53.6k (1h) | $0.12 |
| discussion-summarizer | — | 23.0k | 828 | 13.6k | 9.3k (1h) | $0.05 |

</details>

internal/config/config.go
rianjs: Real catch — fixed: `LoadConfig` now attempts the new-dir read unconditionally (so callers get OAuthClientPath / GrantedScopes etc. alongside the conflict). When `relErr` is set we suppress a hard parse-error propagation so `LoadConfigForRuntime` can still soft-degrade, but we no longer skip the read entirely and return an empty cfg.

internal/config/relocate.go
rianjs: Fixed: `oauthClientPathEquiv` now normalizes both sides via `ExpandPath` before comparison, so `~/oauth_client.json` and the same path fully expanded no longer false-positive as divergent.

internal/cache/cache.go
rianjs: Fixed: `Cache.Clear()` now scopes to `<Root>/<InstanceKey>` instead of the whole tool-level Root, so a future multi-instance Locator can't have one instance's Clear silently evict the others. Test updated.

internal/config/config.go
monit-reviewer: **Settled**: monit-reviewer flagged that LoadConfig never read the new-dir config when a relocation conflict existed, causing callers to receive an empty Config with silent auth and scope-check failures. rianjs fixed it by making the new-dir read unconditional and suppressing hard parse errors on conflict so callers can still soft-degrade.

internal/config/relocate.go
monit-reviewer: **Settled**: monit-reviewer noted that oauthClientPathEquiv compared raw string values without path normalization, causing false ErrRelocationConflict when one side used a tilde and the other an expanded path. rianjs fixed it by normalizing both sides via ExpandPath before comparison.

internal/cache/cache.go
monit-reviewer: **Settled**: monit-reviewer flagged that Cache.Clear() removed the entire tool-level cache root rather than just the current instance's subtree, risking silent eviction of other instances in a future multi-instance scenario. rianjs fixed it by scoping the removal to Root/InstanceKey and updating the test.

internal/config/relocate.go
```diff
@@ -0,0 +1,350 @@
...
+	default:
+		if uerr := yaml.Unmarshal(data, &cfg); uerr != nil {
+			return Config{}, uerr
+		}
+	}
+	return cfg, nil
+}
+
+// configsMaterialEqual compares the user-meaningful subset of two Configs.
+// CredentialRef / GrantedScopes / Keyring.Backend are real user settings and
```
monit-reviewer, line 182: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): When oldPresent && !newPresent and the old config is malformed, detectRelocation returns Kind: relocBothDivergent rather than a distinct kind. LoadConfig's read-priority logic only special-cases relocOldOnly for reading the old file; relocBothDivergent is not in that branch, so a user with a malformed old-only config ends up with all-defaults returned by LoadConfigForRuntime instead of a clear parse error surfaced to the user. A dedicated relocMalformedOld kind (or relocOldOnly with CopyNeeded:false and the parse error) would make the behavior both correct and self-documenting.

internal/config/relocate_test.go
```diff
@@ -0,0 +1,478 @@
...
+
+func TestRelocate_OldOnlyConfigJSON_TriggersCopy(t *testing.T) {
+	// A pre-MON-5371 install on macOS may have only legacy config.json (not
+	// yet promoted to config.yml). Detection must still classify as
+	// relocOldOnly so init copies the JSON into the new dir, where the
+	// existing promoteLegacyConfigJSON migrator picks it up.
+	oldDir, newDir := reloctest(t)
+	if err := os.MkdirAll(oldDir, 0o700); err != nil {
+		t.Fatal(err)
+	}
```
monit-reviewer, line 287: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): TestRelocate_MalformedOld_FailLoud seeds both old and new dirs with config files (malformed old, valid new), so it exercises the both-present branch in detectRelocation. The old-only-malformed code path (oldPresent && !newPresent with invalid YAML) is handled in detectRelocation but has no dedicated test variant. The acceptance matrix lists 'malformed old' as a covered case, but this path is untested.

internal/cache/cache.go
```diff
@@ -1,196 +1,175 @@
...
+	// same subdir under the OLD hand-rolled config dir, which is where a
+	// pure pre-MON-5371 install lived. Both are byte-carried best-effort;
+	// any failure abandons the migration without touching New's result. On
+	// Linux the two paths collapse, so the second probe is a stat-only no-op.
+	probes := []func() (string, error){
+		config.LegacyCacheDir,
+		config.OldHandRolledLegacyCacheDir,
+	}
+	tried := map[string]bool{}
+	for _, p := range probes {
```
monit-reviewer, line 88: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): GetDrives catches json.SyntaxError and json.UnmarshalTypeError via errors.As as cache misses, but not json.InvalidUnmarshalError or io.ErrUnexpectedEOF (which cli-common/cache.ReadResource may wrap around a truncated read). A partially-written cache file producing one of these unwrapped error types would propagate to the caller instead of being demoted to a miss, breaking the 'corrupt → miss' contract. Consider widening the catch or adding a comment that names the specific cli-common wrapping guarantees this relies on.

internal/cache/cache.go
```diff
@@ -1,196 +1,175 @@
...
-		}
-	}
-
-	return status, nil
+	return os.RemoveAll(filepath.Join(c.loc.Root, c.loc.InstanceKey))
 }
 
-// GetDir returns the cache directory path
+// GetDir returns the cache directory path.
 func (c *Cache) GetDir() string {
```
monit-reviewer, line 170: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): GetDir() returns c.loc.Root (the tool-level cache root: <cachedir>/google-readonly), not the per-instance directory where files are actually written (<Root>/<InstanceKey>). TestCache_Clear confirms real data lives one level deeper than GetDir() reports. A future caller using GetDir() to construct data paths will look in the wrong place. Either return filepath.Join(c.loc.Root, c.loc.InstanceKey) or rename to GetRootDir and add a doc comment clarifying this is a tool-root accessor.

internal/config/config.go
```diff
@@ -226,33 +219,98 @@ func ShortenPath(path string) string {
...
+	// when relErr is set (a detect-time ErrRelocationConflict) so callers
+	// soft-degrading via LoadConfigForRuntime actually get the user's new-
+	// dir settings (non-default OAuthClientPath, recorded GrantedScopes,
+	// etc.) — returning all-defaults instead would silently mask issues
+	// like a re-auth requirement or break OAuth client resolution. On a
+	// parse error while relErr is set we DON'T re-surface the bare parse
+	// error: the wrapped conflict already names both paths and the cause,
+	// and a hard-fail here would defeat the soft-degrade contract.
+	cfg := &Config{}
+	read := false
```
monit-reviewer, line 258: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): Double-failure path: when relErr is ErrRelocationConflict and the new config's YAML is also malformed, LoadConfig resets cfg to &Config{} and returns (empty-defaults, relErr). LoadConfigForRuntime then swallows relErr and returns those empty defaults with nil error. Downstream callers receive a wrong OAuthClientPath (empty → default, possibly non-existent) and empty GrantedScopes with no error surfaced. The code comment acknowledges the trade-off, but the consequence — silently running with no config on a double-failure — merits an explicit note on LoadConfigForRuntime's doc comment so callers know this scenario exists.

internal/config/config.go
```diff
@@ -226,33 +219,98 @@ func ShortenPath(path string) string {
 	return path
```
monit-reviewer, line 219: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): LoadConfig returns a non-nil *Config alongside a non-nil error (ErrRelocationConflict), intentionally breaking Go's convention that when err != nil the primary return value is unusable. Future callers of the exported LoadConfig who follow the standard `if err != nil { return err }` pattern will silently discard the valid config and lose the soft-degrade semantics. Making LoadConfig unexported and exposing only LoadConfigForRuntime as the public runtime entry point would eliminate this footgun.

monit-reviewer (approved):
## Automated PR Review



**Reviewed commit:** `94f60bb` | **Previous:** `36c1355` (incremental)

> **Approved** with 6 non-blocking suggestions below. Address at your discretion.

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:harness-enforcement-reviewer | 2 |
| harness-engineering:harness-knowledge-reviewer | 2 |

<details>

**💡 Suggestion** - `internal/cache/cache.go:88`
> GetDrives catches json.SyntaxError and json.UnmarshalTypeError via errors.As as cache misses, but not json.InvalidUnmarshalError or io.ErrUnexpectedEOF (which cli-common/cache.ReadResource may wrap around a truncated read). A partially-written cache file producing one of these unwrapped error types would propagate to the caller instead of being demoted to a miss, breaking the 'corrupt → miss' contract. Consider widening the catch or adding a comment that names the specific cli-common wrapping guarantees this relies on.

</details>

<details>
<summary><strong>harness-engineering:harness-enforcement-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/config/relocate_test.go:287`
> TestRelocate_MalformedOld_FailLoud seeds both old and new dirs with config files (malformed old, valid new), so it exercises the both-present branch in detectRelocation. The old-only-malformed code path (oldPresent && !newPresent with invalid YAML) is handled in detectRelocation but has no dedicated test variant. The acceptance matrix lists 'malformed old' as a covered case, but this path is untested.

**💡 Suggestion** - `internal/config/config.go:219`
> LoadConfig returns a non-nil *Config alongside a non-nil error (ErrRelocationConflict), intentionally breaking Go's convention that when err != nil the primary return value is unusable. Future callers of the exported LoadConfig who follow the standard `if err != nil { return err }` pattern will silently discard the valid config and lose the soft-degrade semantics. Making LoadConfig unexported and exposing only LoadConfigForRuntime as the public runtime entry point would eliminate this footgun.

</details>

<details>
<summary><strong>harness-engineering:harness-knowledge-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/config/relocate.go:182`
> When oldPresent && !newPresent and the old config is malformed, detectRelocation returns Kind: relocBothDivergent rather than a distinct kind. LoadConfig's read-priority logic only special-cases relocOldOnly for reading the old file; relocBothDivergent is not in that branch, so a user with a malformed old-only config ends up with all-defaults returned by LoadConfigForRuntime instead of a clear parse error surfaced to the user. A dedicated relocMalformedOld kind (or relocOldOnly with CopyNeeded:false and the parse error) would make the behavior both correct and self-documenting.

**💡 Suggestion** - `internal/cache/cache.go:170`
> GetDir() returns c.loc.Root (the tool-level cache root: <cachedir>/google-readonly), not the per-instance directory where files are actually written (<Root>/<InstanceKey>). TestCache_Clear confirms real data lives one level deeper than GetDir() reports. A future caller using GetDir() to construct data paths will look in the wrong place. Either return filepath.Join(c.loc.Root, c.loc.InstanceKey) or rename to GetRootDir and add a doc comment clarifying this is a tool-root accessor.

</details>

<details>

**💡 Suggestion** - `internal/config/config.go:258`
> Double-failure path: when relErr is ErrRelocationConflict and the new config's YAML is also malformed, LoadConfig resets cfg to &Config{} and returns (empty-defaults, relErr). LoadConfigForRuntime then swallows relErr and returns those empty defaults with nil error. Downstream callers receive a wrong OAuthClientPath (empty → default, possibly non-existent) and empty GrantedScopes with no error surfaced. The code comment acknowledges the trade-off, but the consequence — silently running with no config on a double-failure — merits an explicit note on LoadConfigForRuntime's doc comment so callers know this scenario exists.

</details>

*3 info-level observations excluded. Run with `--verbose` to include.*

*15 PR discussion threads considered.*

---
<details>
<summary>Completed in 6m 15s | $1.97 | sonnet | daemon 0.2.120 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Mode | Re-review · Cycle 3 · Session resumed |
| Reviewers | hybrid-synthesis, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 6m 15s wall · 16m 02s compute (Reviewers: 4m 14s · Synthesis: 1m 47s) |
| Cost | $1.97 |
| Tokens | 342.4k in / 53.1k out |
| Turns | 6 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 52.2k | 5.7k | 13.6k | 38.5k (1h) | $0.23 |
| harness-engineering:harness-architecture-reviewer | sonnet | 53.9k | 5.7k | 2.1k | 51.8k (1h) | $0.33 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 53.9k | 13.0k | 2.1k | 51.8k (1h) | $0.44 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 53.9k | 10.4k | 2.1k | 51.8k (1h) | $0.40 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 53.3k | 10.4k | 2.1k | 51.1k (1h) | $0.39 |
| security:security-code-auditor | haiku | 53.5k | 7.6k | 0 | 53.5k (1h) | $0.15 |
| discussion-summarizer | — | 21.7k | 345 | 13.6k | 8.1k (1h) | $0.04 |

</details>

rianjs:
### Findings

**Blocker:** malformed durable config still soft-degrades to defaults on runtime paths.  
`DetectConfigRelocation` correctly returns `ErrRelocationConflict` for malformed old/new configs (`internal/config/relocate.go:131-149`), but `LoadConfig` now suppresses canonical read/parse failures whenever `relErr` is set and leaves `cfg` empty (`internal/config/config.go:262-279`). `LoadConfigForRuntime` then suppresses every `ErrRelocationConflict` and returns that defaulted config with `nil` error (`internal/config/relocate.go:335-340`). So a malformed new config, or old-only malformed config, becomes “warn + use defaults” for non-init commands. That violates the durable-state §3.2 malformed-old/malformed-new fail-loud contract and can silently switch `credential_ref` back to `google-readonly/default`.

I’d distinguish parse/corrupt relocation errors from parseable old/new divergence, or only soft-suppress when the canonical new config was successfully read.

rianjs:
No findings. The final changes line up with the intended split: init remains fail-loud and mutation-gated, runtime only soft-degrades when it has a successfully read canonical config, malformed canonical config hard-fails, token handling stays in the §1.8 migrator, and cache behavior remains disposable and scoped.

Focused verification passed: `go test ./internal/config ./internal/cache ./internal/keychain`.

PR #138, opened 2026-05-21
fix(config): clear --all removes legacy config file, not just legacy cache

### LINKED ISSUE #137: config clear --all removes the legacy Drive cache but not the legacy config dir
## Symptom

After the platform-config-dir migration ([#134](https://github.com/open-cli-collective/google-readonly/pull/134)), a user with a pre-MON-5371 install on macOS or Windows has files in **both** the legacy `~/.config/google-readonly/` directory and the new `os.UserConfigDir/google-readonly/` directory.

Running `gro config clear --all` followed by `gro init` does not get them to a clean state:

1. `clear --all` removes the keychain token, the new-location `config.yml`, the new-location Drive cache, and the legacy Drive cache.
2. It does **not** remove the legacy config file (`~/.config/google-readonly/config.json` or `config.yml`).
3. On the next `gro init`, `DetectConfigRelocation` sees the legacy file is still present, the new dir is missing (we just removed it), so it returns `relocOldOnly` with `CopyNeeded=true`, then `ApplyConfigRelocation` copies the legacy file forward.
4. If the user has independently re-created the new-location file (via, say, an installer that drives `gro init` in a loop or a partial intermediate run), step 3 instead returns `relocBothDivergent` and `gro init` refuses to proceed:

   > warning: config: shared old/new config diverge: old /Users/x/.config/google-readonly/config.json and new /Users/x/Library/Application Support/google-readonly/config.yml have different settings; reconcile (or delete one) before running gro init

## Why this looks like an oversight

`runClear --all` in [`internal/cmd/config/config.go`](https://github.com/open-cli-collective/google-readonly/blob/main/internal/cmd/config/config.go#L259-L284) explicitly handles **both** the new and legacy *cache* dirs:

```go
if cacheDir, cerr := config.CacheDirPath(); cerr == nil {
    if rmErr := os.RemoveAll(cacheDir); rmErr != nil {
        return fmt.Errorf("removing Drive metadata cache %s: %w", config.ShortenPath(cacheDir), rmErr)
    }
    fmt.Printf("Removed Drive metadata cache at %s.\n", config.ShortenPath(cacheDir))
}
if legacy, lerr := config.LegacyCacheDir(); lerr == nil {
    if rmErr := os.RemoveAll(legacy); rmErr != nil {
        return fmt.Errorf("removing legacy Drive cache %s: %w", config.ShortenPath(legacy), rmErr)
    }
}
```

The comment above this block makes the intent clear:

> Drive metadata cache: an explicit full reset removes **BOTH** the current cache dir and the legacy (pre-B2b) one directly — no `cache.New()`, so no migrate-then-delete dance and no `MkdirAll` side-effect.

But the equivalent block for the config file removes only the new-location `cfgPath`. The legacy config dir's `config.{json,yml}` is never touched. The asymmetry between cache (both removed) and config (only new removed) reads as the config-side cleanup not being updated when MON-5371 introduced platform-specific config dirs (the cache-side cleanup landed in B2b / [#130](https://github.com/open-cli-collective/google-readonly/pull/130); MON-5371 landed in [#134](https://github.com/open-cli-collective/google-readonly/pull/134) but didn't extend `clear --all`).

I think the `ApplyConfigRelocation` copy-leave-old policy is the right default for `init` — the legacy file is a recovery point for the user. But for an explicit `clear --all` ("Also remove config.yml and the Drive metadata cache"), the user is asking for a full reset; leaving the legacy file behind defeats the operation and silently re-creates the divergence trap on the next init.

## Proposed fix

In `runClear`, after `os.Remove(cfgPath)`, also remove the legacy config file. Something like:

```go
if legacyDir, lerr := config.LegacyConfigDir(); lerr == nil {
    for _, name := range []string{config.ConfigFileYAML, config.ConfigFile} {
        legacyPath := filepath.Join(legacyDir, name)
        switch err := os.Remove(legacyPath); {
        case err == nil:
            fmt.Printf("Removed legacy %s.\n", config.ShortenPath(legacyPath))
        case os.IsNotExist(err):
            // fine
        default:
            return fmt.Errorf("removing legacy %s: %w", config.ShortenPath(legacyPath), err)
        }
    }
}
```

Notes:
- The OAuth client JSON in the legacy dir is still preserved (consistent with the existing "never removed" guarantee for deployment material).
- A `LegacyConfigDir()` helper would need to exist or be added; the package already has `LegacyCacheDir()` as precedent.
- `--dry-run` would need parallel updates so users see what `--all` will actually remove before running it.

## Impact today

Downstream installers ([`claude-desktop-mcp`](https://github.com/SignalFT/claude-desktop-mcp) PRs [#31](https://github.com/SignalFT/claude-desktop-mcp/pull/31) Windows + #32 macOS) currently work around this by running `gro config clear --all` and then `os.unlink`-ing the legacy `config.json` themselves before invoking `gro init`. Moving the cleanup into gro itself would let downstream installers drop that workaround.

## Repro

On a fresh macOS box:
1. Seed `~/.config/google-readonly/config.json` with `{}` and `~/Library/Application Support/google-readonly/config.yml` with `credential_ref: foo/bar`.
2. `gro config clear --all` — succeeds.
3. `ls ~/.config/google-readonly/` — `config.json` is still there.
4. Touch `~/Library/Application Support/google-readonly/config.yml` with anything → `gro init` → divergence warning.

Happy to PR if useful.

rianjs:
## Summary

`gro config clear --all` removed the legacy Drive cache (via `LegacyCacheDir`) but did NOT remove the legacy config file at `<old-hand-rolled-dir>/config.{yml,json}`. After MON-5371 (#134) relocated the config dir to OS-native paths on macOS/Windows, this asymmetry re-armed the `DetectConfigRelocation` divergence trap on the next `gro init`. (The cache side got the both-dir treatment in B2b / #130; the config side was missed when #134 introduced the relocation.)

This change mirrors the cache-side symmetry on the config side and hardens `--all` against the malformed state it exists to wipe.

## Changes

- **`config.OldHandRolledConfigDir`** — exported primitive, symmetric with the existing `OldHandRolledTokenPath` / `LegacyCacheDir`.
- **`config.GetConfigDirNoCreate`** — symmetric with `GetConfigPathNoCreate`; lets the candidate helper resolve without side-effects.
- **`internal/cmd/config.configFilesForClear`** — per-file deduped candidate list covering both canonical `config.{yml,json}` and the legacy hand-rolled dir's `config.{yml,json}` (canonical `.json` is included because `LoadConfig` reads it when `.yml` is absent). `configFilesForClearFn` is a package-var test seam so Linux CI can exercise the macOS/Windows divergent-dir branch.
- **`runClear` hardening** (§6.6 pattern 7) — under `--all`, an `OpenNoMigrate` failure (e.g. invalid `keyring.backend` from a malformed canonical config) soft-degrades with a warning so the file scrub still completes. Plain `clear` (no `--all`) still hard-fails on a broken keyring; the user didn't ask for a wipe there.

OAuth client JSON is intentionally NOT removed (deployment material exclusion preserved per the existing trailing note).

## Tests

`internal/cmd/config/clear_legacy_test.go`:

- `TestConfigClearAll_RemovesAllConfigCandidates` — seeds all four candidate files via the synthetic-distinct-dir seam; asserts all removed.
- `TestConfigClearAll_DryRunReportsAllCandidates` — dry-run names every candidate; files still on disk.
- `TestConfigClearAll_LegacyAbsent_OnlyCanonical` — absent legacy files don't produce spurious "Removed" lines.
- `TestConfigFilesForClear_PathIdentityDedup` — per-file dedup pin; also pins canonical `config.json` inclusion (Codex r2 regression).
- `TestConfigClearAll_MalformedCanonicalConfig_StillScrubsFiles` — §6.6 pattern 7 recovery-contract pin (matches nrq's `TestConfigClear_All_MalformedConfig_StillScrubsFiles`).
- `TestConfigClear_PlainStillHardFailsOnBrokenKeyring` — non-`--all` path stays strict.

`make check` passes locally: 1384 tests across 32 packages, 0 lint issues.

## Test plan

- [x] `make check` green
- [x] Codex architect plan review converged (blockers=0 majors=0 minors=0 nits=0 on r3)
- [ ] Manual repro from issue#137 "Repro" section no longer reproduces (will verify in CI environment)

Closes #137
[MON-5433]

### DIFF
--- a/internal/cmd/config/clear_legacy_test.go
+++ b/internal/cmd/config/clear_legacy_test.go
@@ -1,0 +1,215 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/open-cli-collective/google-readonly/internal/credtest"
+)
+
+// withSyntheticConfigCandidates overrides configFilesForClearFn so Linux CI
+// can exercise the macOS/Windows "old != new" branch with explicit paths.
+// Same shape as MON-5373 nrq's credentialFileCandidates seam.
+func withSyntheticConfigCandidates(t *testing.T, paths []string) {
+	t.Helper()
+	orig := configFilesForClearFn
+	configFilesForClearFn = func() ([]string, error) { return paths, nil }
+	t.Cleanup(func() { configFilesForClearFn = orig })
+}
+
+func seedFile(t *testing.T, path, contents string) {
+	t.Helper()
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+		t.Fatalf("MkdirAll %s: %v", filepath.Dir(path), err)
+	}
+	if err := os.WriteFile(path, []byte(contents), 0o600); err != nil {
+		t.Fatalf("WriteFile %s: %v", path, err)
+	}
+}
+
+func TestConfigClearAll_RemovesAllConfigCandidates(t *testing.T) {
+	credtest.Setup(t)
+	tmp := t.TempDir()
+	newYML := filepath.Join(tmp, "new", "config.yml")
+	newJSON := filepath.Join(tmp, "new", "config.json")
+	oldYML := filepath.Join(tmp, "old", "config.yml")
+	oldJSON := filepath.Join(tmp, "old", "config.json")
+	for _, p := range []string{newYML, newJSON, oldYML, oldJSON} {
+		seedFile(t, p, "credential_ref: foo/bar\n")
+	}
+	withSyntheticConfigCandidates(t, []string{newYML, newJSON, oldYML, oldJSON})
+
+	_ = capture(t, func() {
+		if err := runClear(true, false); err != nil {
+			t.Fatalf("runClear: %v", err)
+		}
+	})
+
+	for _, p := range []string{newYML, newJSON, oldYML, oldJSON} {
+		if _, err := os.Stat(p); !os.IsNotExist(err) {
+			t.Errorf("--all must remove %s (stat err=%v)", p, err)
+		}
+	}
+}
+
+func TestConfigClearAll_DryRunReportsAllCandidates(t *testing.T) {
+	credtest.Setup(t)
+	tmp := t.TempDir()
+	newYML := filepath.Join(tmp, "new", "config.yml")
+	oldJSON := filepath.Join(tmp, "old", "config.json")
+	seedFile(t, newYML, "credential_ref: foo/bar\n")
+	seedFile(t, oldJSON, `{"credential_ref":"legacy/bar"}`)
+	withSyntheticConfigCandidates(t, []string{newYML, oldJSON})
+
+	out := capture(t, func() {
+		if err := runClear(true, true); err != nil {
+			t.Fatalf("runClear dry-run: %v", err)
+		}
+	})
+
+	for _, p := range []string{newYML, oldJSON} {
+		if !strings.Contains(out, p) && !strings.Contains(out, shortened(t, p)) {
+			t.Errorf("--dry-run --all should name %q in output, got:\n%s", p, out)
+		}
+		if _, err := os.Stat(p); err != nil {
+			t.Errorf("--dry-run must not remove %s (stat err=%v)", p, err)
+		}
+	}
+}
+
+func TestConfigClearAll_LegacyAbsent_OnlyCanonical(t *testing.T) {
+	credtest.Setup(t)
+	tmp := t.TempDir()
+	newYML := filepath.Join(tmp, "new", "config.yml")
+	oldYML := filepath.Join(tmp, "old", "config.yml") // never seeded
+	seedFile(t, newYML, "credential_ref: foo/bar\n")
+	withSyntheticConfigCandidates(t, []string{newYML, oldYML})
+
+	out := capture(t, func() {
+		if err := runClear(true, false); err != nil {
+			t.Fatalf("runClear: %v", err)
+		}
+	})
+
+	if _, err := os.Stat(newYML); !os.IsNotExist(err) {
+		t.Errorf("--all must remove canonical %s (stat err=%v)", newYML, err)
+	}
+	// Absent legacy must not produce a "Removed" line. We accept silence on
+	// IsNotExist.
+	if strings.Contains(out, oldYML) {
+		t.Errorf("absent %s should not appear in --all output, got:\n%s", oldYML, out)
+	}
+}
+
+func TestConfigFilesForClear_PathIdentityDedup(t *testing.T) {
+	credtest.Setup(t)
+	// Engineer oldDir == newDir by pointing XDG_CONFIG_HOME at the same root
+	// the canonical resolver uses. credtest.Setup put us under a hermetic
+	// statedirtest env; on Linux that already produces old==new — the test
+	// is meaningful there. On macOS/Windows the resolvers diverge by design
+	// and this assertion just verifies no spurious duplication when they
+	// happen to coincide (the canonical paths are themselves distinct).
+	paths, err := configFilesForClear()
+	if err != nil {
+		t.Fatalf("configFilesForClear: %v", err)
+	}
+	seen := map[string]int{}
+	for _, p := range paths {
+		seen[p]++
+	}
+	for p, n := range seen {
+		if n != 1 {
+			t.Errorf("path %s appeared %d times; want 1 (per-file dedup)", p, n)
+		}
+	}
+	// Both forms must always be present at the canonical dir (regression pin
+	// for the Codex r2 minor: canonical config.json was originally missed).
+	newDir, derr := newDirFor(t)
+	if derr != nil {
+		t.Fatalf("newDirFor: %v", derr)
+	}
+	wantYML := filepath.Join(newDir, "config.yml")
+	wantJSON := filepath.Join(newDir, "config.json")
+	if seen[wantYML] == 0 {
+		t.Errorf("candidate list missing canonical config.yml at %s; got %v", wantYML, paths)
+	}
+	if seen[wantJSON] == 0 {
+		t.Errorf("candidate list missing canonical config.json at %s; got %v", wantJSON, paths)
+	}
+}
+
+func TestConfigClearAll_MalformedCanonicalConfig_StillScrubsFiles(t *testing.T) {
+	credtest.Setup(t)
+	tmp := t.TempDir()
+	newYML := filepath.Join(tmp, "new", "config.yml")
+	// File contents are irrelevant for scrub-completion; the failure mode
+	// being pinned is "keyring open fails / config malformed → file scrub
+	// still completes". We force the keyring side via an unknown backend
+	// env value, which credstore.Open() must reject.
+	seedFile(t, newYML, "credential_ref: foo/bar\n")
+	t.Setenv("GOOGLE_READONLY_KEYRING_BACKEND", "this-backend-does-not-exist")
+	withSyntheticConfigCandidates(t, []string{newYML})
+
+	out := capture(t, func() {
+		// We expect runClear to NOT return an error under --all even though
+		// the keyring open fails — that is the recovery contract.
+		if err := runClear(true, false); err != nil {
+			t.Fatalf("runClear under --all must soft-degrade on keyring failure, got: %v", err)
+		}
+	})
+
+	if _, err := os.Stat(newYML); !os.IsNotExist(err) {
+		t.Errorf("--all must scrub %s even when keyring open fails (stat err=%v)", newYML, err)
+	}
+	// We don't assert exact warning wording, only that a Removed line was
+	// emitted for the candidate (cheap visibility into "did we actually
+	// reach the scrub").
+	if !strings.Contains(out, newYML) && !strings.Contains(out, shortened(t, newYML)) {
+		t.Errorf("expected runClear to print Removed line for %s, got:\n%s", newYML, out)
+	}
+}
+
+func TestConfigClear_PlainStillHardFailsOnBrokenKeyring(t *testing.T) {
+	credtest.Setup(t)
+	t.Setenv("GOOGLE_READONLY_KEYRING_BACKEND", "this-backend-does-not-exist")
+
+	_ = capture(t, func() {
+		err := runClear(false, false)
+		if err == nil {
+			t.Fatal("plain `clear` (no --all) must surface a keyring open failure; got nil")
+		}
+	})
+}
+
+// shortened returns config.ShortenPath(p) via a callback-style indirection
+// that avoids importing internal/config in the test file just for one call.
+// Production code already prints the shortened form, so the test must match
+// whichever form appears.
+func shortened(t *testing.T, p string) string {
+	t.Helper()
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return p
+	}
+	if strings.HasPrefix(p, home) {
+		return "~" + strings.TrimPrefix(p, home)
+	}
+	return p
+}
+
+// newDirFor returns the canonical config dir as configFilesForClear would
+// resolve it (via the live config.GetConfigDirNoCreate). Kept as a helper
+// so tests don't have to import internal/config solely for this.
+func newDirFor(t *testing.T) (string, error) {
+	t.Helper()
+	paths, err := configFilesForClear()
+	if err != nil {
+		return "", err
+	}
+	if len(paths) == 0 {
+		t.Fatal("configFilesForClear returned no paths")
+	}
+	return filepath.Dir(paths[0]), nil
+}
--- a/internal/cmd/config/clear_legacy_test.go
+++ b/internal/cmd/config/clear_legacy_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	appconfig "github.com/open-cli-collective/google-readonly/internal/config"
 	"github.com/open-cli-collective/google-readonly/internal/credtest"
 )
 
@@ -70,7 +71,7 @@ func TestConfigClearAll_DryRunReportsAllCandidates(t *testing.T) {
 	})
 
 	for _, p := range []string{newYML, oldJSON} {
-		if !strings.Contains(out, p) && !strings.Contains(out, shortened(t, p)) {
+		if !strings.Contains(out, p) && !strings.Contains(out, appconfig.ShortenPath(p)) {
 			t.Errorf("--dry-run --all should name %q in output, got:\n%s", p, out)
 		}
 		if _, err := os.Stat(p); err != nil {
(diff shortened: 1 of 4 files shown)


rianjs:
**Major**
- [internal/cmd/config/config.go:278](internal/cmd/config/config.go:278) still lets a keyring query failure block the `--all` file scrub. The PR soft-degrades only `OpenNoMigrate()` failure, but if `st.HasToken()` returns an error, `runClear(true, false)` returns before removing any `cfgPaths`. That leaves the legacy config files in place in another broken-keyring state, which conflicts with the §6.6 recovery contract. Under `all`, handle `HasToken()` failure like open failure: warn, skip token cleanup, continue file scrub; keep non-`--all` strict.

**Minor**
- [internal/cmd/config/clear_legacy_test.go:143](internal/cmd/config/clear_legacy_test.go:143) overclaims the malformed-config recovery pin. The test name/comment says “MalformedCanonicalConfig”, but it writes a valid synthetic config file and forces failure via `GOOGLE_READONLY_KEYRING_BACKEND`. Either rename it to a generic `OpenNoMigrateFailure_StillScrubsFiles`, or add a real canonical malformed `config.yml` case so the product-intent claim is actually verified.

**Nit**
- [internal/cmd/config/clear_legacy_test.go:186](internal/cmd/config/clear_legacy_test.go:186) reimplements `config.ShortenPath` in test code. This package already imports `internal/config` elsewhere as `appconfig`; using the production helper avoids a tiny parallel implementation.

rianjs:
## TDD coverage assessment

**Verdict: close to sufficient — one real gap that Codex also flagged.**

The new tests cleanly cover the bug-fix intent (legacy-file scrub, dry-run parity, per-file dedup, canonical `config.json` regression pin) and the §6.6 pattern-7 recovery contract for `OpenNoMigrate` failure. The synthetic-candidates seam is a good fit for Linux CI exercising the macOS/Windows divergent-dir branch.

### Real gap
- **`st.HasToken()` failure under `--all` is untested.** `runClear` soft-degrades `OpenNoMigrate` failure but `HasToken()` returning an error still short-circuits the scrub at `internal/cmd/config/config.go:278`. There's no test where the keyring opens but `HasToken` errors (e.g. partial backend corruption). Add a test that pins this — same soft-degrade should apply, since the contract is "broken state must not block its own wipe." (Codex flagged this independently; not novel here.)

### Minor
- **`TestConfigClearAll_MalformedCanonicalConfig_StillScrubsFiles` is mis-named.** It seeds a valid YAML and forces failure via `GOOGLE_READONLY_KEYRING_BACKEND`. Since `runClear` no longer calls `LoadConfig` (paths come from `configFilesForClearFn`), there is no malformed-canonical path to verify here — the actual contract being pinned is "broken keyring under `--all`." Rename to `TestConfigClearAll_BrokenKeyring_StillScrubsFiles` for honesty, OR add a second case that writes genuinely malformed YAML to verify it's tolerated (it should be — paths are read without parsing — but no test confirms).
- **Stderr warning not asserted.** The soft-degrade emits `warning: could not open keyring (...)` on stderr. Tests don't capture stderr, so user-visibility is unverified.
- **OAuth client JSON exclusion is not pinned by the new tests.** The "deployment material is never removed" claim relies on it not appearing in `configFilesForClear`'s output. The dedup test could cheaply assert `OAuthClientFile` is absent from `paths`.

### Acceptable as-is
- No `os.Remove` error-mid-loop test (hard to portably trigger; the failure mode is documented and the error is surfaced, not swallowed).
- Real-resolver `configFilesForClear` test runs old==new on Linux — that's a known limitation of statedirtest, accepted via the seam.

**Recommend**: add the `HasToken` soft-degrade test (the only real gap) and rename the malformed-config test for honesty. Other items are optional polish.

rianjs:
Addressed Codex r1 + TDD findings in 3d4989a:

- **Codex Major** (HasToken soft-degrade): symmetric soft-degrade now applied. HasToken failure under `--all` emits a warning, eagerly closes the store, drops the reference. Nil-guarded the deferred close.
- **Codex Minor / TDD Minor** (mis-named malformed-config test): split into two honest tests — `BrokenKeyringOpen_StillScrubsFiles` (forces OpenNoMigrate failure) and `MalformedCanonicalConfig_StillScrubsFiles` (genuinely malformed YAML on disk; pins that runClear no longer reads the config). Plain-clear hard-fail test stays.
- **Codex Nit** (parallel ShortenPath): test file now uses `appconfig.ShortenPath` directly.
- **TDD Minor** (OAuth client JSON exclusion): added `TestConfigFilesForClear_ExcludesOAuthClientJSON`.
- **TDD Minor** (stderr warning capture): deferred — keeping the production code minimal; the warning text is non-load-bearing and code review covers the symmetric branch.

`make check` green: 1387 tests, 0 lint issues.

rianjs:
Codex r2: 0 blockers, 0 majors, 1 minor, 0 nits — convergence reached.

Acknowledging the remaining minor: `TestConfigClearAll_MalformedCanonicalConfig_StillScrubsFiles` seeds malformed YAML at a synthetic path, so it doesn't exercise a malformed real-canonical config blocking `keychain.OpenNoMigrate()`. The *production* fact being claimed (runClear no longer reads the canonical config — paths come from `configFilesForClearFn` only) is verified by code inspection rather than this test. Leaving the test as-is per family precedent of buggering off marginal test-naming nits. Moving to daemon pass.

internal/cmd/config/config.go
```diff
@@ -210,33 +250,67 @@ func runTest(cmd *cobra.Command, _ []string) error {
...
+		default:
+			return fmt.Errorf("checking stored token: %w", herr)
+		}
 	}
 
 	if dryRun {
-		if hasTok {
+		switch {
+		case st == nil:
+			fmt.Println("Would remove: (keyring unavailable — token state unknown)")
```
monit-reviewer, line 304: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): Dry-run and live-run diverge on absent candidate files: `--dry-run --all` unconditionally prints 'Would remove: X' for every candidate path returned by `configFilesForClearFn()`, but the live path silently skips non-existent files with no output. A user previewing with dry-run will see removal lines for files the real run would silently skip, making dry-run output unreliable as a preview. Consider checking `os.Stat` before printing, or using 'Would remove (if present):' phrasing for candidates that may not exist.

internal/cmd/config/config.go
```diff
@@ -247,23 +321,28 @@ func runClear(all, dryRun bool) error {
...
 
 	if all {
-		switch err := os.Remove(cfgPath); {
-		case err == nil:
-			fmt.Printf("Removed %s.\n", config.ShortenPath(cfgPath))
-		case os.IsNotExist(err):
-			fmt.Printf("No %s to remove.\n", config.ShortenPath(cfgPath))
-		default:
-			return fmt.Errorf("removing %s: %w", config.ShortenPath(cfgPath), err)
+		for _, p := range cfgPaths {
```
monit-reviewer, line 337: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): The `os.IsNotExist` branch is now silent. The old code printed 'No X to remove.' for absent files; the new multi-candidate code skips with no output. On an already-clean system (the common idempotent re-run case), the user gets zero confirmation that any config-file candidates were checked. A single summary line such as 'No legacy config files found.' when all candidates are absent would preserve the UX parity the old code had.

internal/cmd/config/config.go
```diff
@@ -18,6 +19,45 @@ import (
...
+// and old hand-rolled dirs because LoadConfig still reads new-dir
+// config.json when YAML is absent, and the relocation gate can still copy
+// either form forward from the old dir. The OAuth client JSON is
+// intentionally excluded (deployment material, per the existing trailing
+// note in runClear). Resolves WITHOUT creating directories.
+func configFilesForClear() ([]string, error) {
+	newDir, err := config.GetConfigDirNoCreate()
+	if err != nil {
+		return nil, err
+	}
```
monit-reviewer, line 35: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): `OldHandRolledConfigDir()` error is silently discarded (`oldDir, _ := ...`). The comment documents 'best-effort: no HOME → skip legacy', but any unexpected `os.UserHomeDir` error will also silently omit legacy scrub candidates with no user-visible feedback. Under `--all` (a recovery path), silent partial cleanup could leave the user in a confusing half-wiped state. A `fmt.Fprintf(os.Stderr, ...)` warning on non-nil err — symmetric with the `OpenNoMigrate` soft-degrade pattern used elsewhere in the same function — would make degradation visible.

internal/cmd/config/clear_legacy_test.go
```diff
@@ -0,0 +1,232 @@
...
+func TestConfigClear_PlainStillHardFailsOnBrokenKeyring(t *testing.T) {
+	credtest.Setup(t)
+	t.Setenv("GOOGLE_READONLY_KEYRING_BACKEND", "this-backend-does-not-exist")
+
+	_ = capture(t, func() {
+		err := runClear(false, false)
+		if err == nil {
+			t.Fatal("plain `clear` (no --all) must surface a keyring open failure; got nil")
+		}
+	})
```
monit-reviewer, line 216: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): `newDirFor` derives the canonical config dir by calling `configFilesForClear()` and taking `filepath.Dir(paths[0])`, silently assuming canonical paths are always returned first. This is an implicit ordering contract on a private implementation detail with no enforcement. If `configFilesForClear()` is ever reordered, `TestConfigFilesForClear_PathIdentityDedup` will silently test the wrong directory without a diagnostic error. A direct call to `config.GetConfigDirNoCreate()` (already exported by this PR) would make the intent explicit and eliminate the fragility.

monit-reviewer (approved):
## Automated PR Review



**Reviewed commit:** `3d4989a`

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:harness-knowledge-reviewer | 2 |

<details>

**💡 Suggestion** - `internal/cmd/config/config.go:304`
> Dry-run and live-run diverge on absent candidate files: `--dry-run --all` unconditionally prints 'Would remove: X' for every candidate path returned by `configFilesForClearFn()`, but the live path silently skips non-existent files with no output. A user previewing with dry-run will see removal lines for files the real run would silently skip, making dry-run output unreliable as a preview. Consider checking `os.Stat` before printing, or using 'Would remove (if present):' phrasing for candidates that may not exist.

</details>

<details>
<summary><strong>harness-engineering:harness-knowledge-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/cmd/config/config.go:337`
> The `os.IsNotExist` branch is now silent. The old code printed 'No X to remove.' for absent files; the new multi-candidate code skips with no output. On an already-clean system (the common idempotent re-run case), the user gets zero confirmation that any config-file candidates were checked. A single summary line such as 'No legacy config files found.' when all candidates are absent would preserve the UX parity the old code had.

**💡 Suggestion** - `internal/cmd/config/config.go:35`
> `OldHandRolledConfigDir()` error is silently discarded (`oldDir, _ := ...`). The comment documents 'best-effort: no HOME → skip legacy', but any unexpected `os.UserHomeDir` error will also silently omit legacy scrub candidates with no user-visible feedback. Under `--all` (a recovery path), silent partial cleanup could leave the user in a confusing half-wiped state. A `fmt.Fprintf(os.Stderr, ...)` warning on non-nil err — symmetric with the `OpenNoMigrate` soft-degrade pattern used elsewhere in the same function — would make degradation visible.

</details>

<details>

**💡 Suggestion** - `internal/cmd/config/clear_legacy_test.go:216`
> `newDirFor` derives the canonical config dir by calling `configFilesForClear()` and taking `filepath.Dir(paths[0])`, silently assuming canonical paths are always returned first. This is an implicit ordering contract on a private implementation detail with no enforcement. If `configFilesForClear()` is ever reordered, `TestConfigFilesForClear_PathIdentityDedup` will silently test the wrong directory without a diagnostic error. A direct call to `config.GetConfigDirNoCreate()` (already exported by this PR) would make the intent explicit and eliminate the fragility.

</details>

---
<details>
<summary>Completed in 2m 30s | $0.73 | sonnet | daemon 0.2.120 | Glorfindel</summary>

| Field | Value |
|---|---|
| Model | sonnet |
| Reviewers | hybrid-synthesis, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude · sonnet |
| Duration | 2m 30s wall · 6m 14s compute (Reviewers: 1m 10s · Synthesis: 1m 16s) |
| Cost | $0.73 |
| Tokens | 118.4k in / 22.8k out |
| Turns | 6 |

**Per-workstream usage**

|---|---|---|---|---|---|---|
| hybrid-synthesis | sonnet | 31.7k | 4.7k | 0 | 31.7k (1h) | $0.20 |
| harness-engineering:harness-architecture-reviewer | sonnet | 17.3k | 3.0k | 2.1k | 15.1k (1h) | $0.11 |
| harness-engineering:harness-enforcement-reviewer | sonnet | 17.3k | 3.6k | 2.1k | 15.1k (1h) | $0.12 |
| harness-engineering:harness-knowledge-reviewer | sonnet | 17.3k | 3.9k | 2.1k | 15.1k (1h) | $0.12 |
| harness-engineering:harness-self-documenting-code-reviewer | sonnet | 17.3k | 4.1k | 2.1k | 15.1k (1h) | $0.13 |
| security:security-code-auditor | haiku | 17.5k | 3.6k | 0 | 17.5k (1h) | $0.05 |

</details>

internal/cmd/config/config.go
rianjs: Acknowledged as low/marginal — keeping as-is. The dry-run/live print divergence is intentional (dry-run answers 'what would be removed' speculatively; live confirms what actually happened). Per the family precedent (`[[feedback-pr-review-loop]]`) we bugger off marginal nits to avoid review-loop churn on already-approved PRs.

internal/cmd/config/config.go
rianjs: Acknowledged as low/marginal — keeping as-is. The dry-run/live print divergence is intentional (dry-run answers 'what would be removed' speculatively; live confirms what actually happened). Per the family precedent (`[[feedback-pr-review-loop]]`) we bugger off marginal nits to avoid review-loop churn on already-approved PRs.

internal/cmd/config/config.go
rianjs: Acknowledged as low/marginal — keeping as-is. The dry-run/live print divergence is intentional (dry-run answers 'what would be removed' speculatively; live confirms what actually happened). Per the family precedent (`[[feedback-pr-review-loop]]`) we bugger off marginal nits to avoid review-loop churn on already-approved PRs.

internal/cmd/config/clear_legacy_test.go
rianjs: Acknowledged as low/marginal — keeping as-is. The dry-run/live print divergence is intentional (dry-run answers 'what would be removed' speculatively; live confirms what actually happened). Per the family precedent (`[[feedback-pr-review-loop]]`) we bugger off marginal nits to avoid review-loop churn on already-approved PRs.

